PR #332, opened 2026-02-10
test: Add unit tests for MCP server handlers

### LINKED ISSUE #187: Add unit tests for MCP server handlers
## Task

Add unit tests for MCP server handlers in `mcp/handlers_test.go`

## What to test

6 handlers in `mcp/handlers.go`:
- `HandleAnalyzeCode`
- `HandleCheckComplexity`
- `HandleDetectClones`
- `HandleCheckCoupling`
- `HandleFindDeadCode`
- `HandleGetHealthScore`

## Example

```go
func TestHandleAnalyzeCode(t *testing.T) {
    deps := &Dependencies{/* ... */}
    request := mcp.CallToolRequest{
        Params: mcp.CallToolRequestParams{
            Arguments: map[string]interface{}{
                "path": "testdata/python/simple",
            },
        },
    }
    
    result, err := HandleAnalyzeCode(context.Background(), request, deps)
    if err != nil {
        t.Errorf("unexpected error: %v", err)
    }
    // Verify result...
}
```

## Tips

- Use test data in `testdata/python/simple/`
- Test both success and error cases
- Check existing tests: `internal/analyzer/*/\*_test.go`
- Run: `go test ./mcp/...`

## Acceptance

- [ ] `mcp/handlers_test.go` created
- [ ] Tests for all 6 handlers
- [ ] Tests pass with `go test ./mcp/...`
- [ ] Coverage > 70%

BenjaminGoehry:
Closes #187 

### Summary
These tests validate the behavior of all `HandlerSet `MCP tool handlers, ensuring they handle input validation, errors, and output formatting correctly. The focus is on handler behavior rather than internal analysis logic.

Handlers Tested:
- `HandleAnalyzeCode`
  - Invalid arguments: Non-map arguments trigger an error with the correct message.
  - Missing path: Omitting the "path" parameter triggers an error.
  - Non-existent path: Returns an error when the provided path does not exist.
  - Success: Valid Python file input produces JSON output containing "health_score".
  - Full output mode: "output_mode": "full" returns detailed JSON content.
  - Analyses filter: "analyses": ["complexity"] executes only the specified analysis.

- `HandleCheckComplexity`
  - Invalid arguments: Non-map arguments trigger an error.
  - Missing path: Absence of "path" triggers an error.
  - Non-existent path: Returns an error.
  - Success: Valid Python file input returns results; ensures non-empty content.

- `HandleDetectClones`
  - Invalid arguments: Non-map arguments trigger an error.
  - Missing path: Triggers an error.
  - Non-existent path: Returns an error.
  - Success (summary mode): Returns results for a valid file path.
  - Success (detailed mode): Returns results when "output_mode": "detailed".
  - Analyses filter: "analyses": ["clone"] executes only clone analysis.

- `HandleCheckCoupling`
  - Invalid arguments: Non-map arguments trigger an error.
  - Missing path: Returns an error.
  - Non-existent path: Returns an error.
  - Success (summary & full): Valid Python file input returns results; summary mode includes a "summary" key.

- `HandleFindDeadCode`
  - Invalid arguments: Non-map arguments trigger an error.
  - Missing path: Returns an error.
  - Non-existent path: Returns an error.
  - Success: Valid Python file input with "min_severity" parameter returns results containing "summary".
  - Min severity variations: Tests "warning", "info", "critical", and default severity.

- `HandleGetHealthScore`
  - Invalid arguments: Non-map arguments trigger an error.
  - Missing path: Returns an error.
  - Non-existent path: Returns an error.
  - Success: Valid Python file input returns JSON output containing "health_score", "grade", and "category_scores".

Capture d&#39;écran 2026-02-10 205424

### DIFF
--- a/mcp/handlers_test.go
+++ b/mcp/handlers_test.go
@@ -260,9 +260,7 @@ func TestHandleCheckComplexity(t *testing.T) {
 }
 
 func TestHandleCheckCoupling(t *testing.T) {
-
 	errTrue := true
-	errFalse := false
 
 	tests := map[string]struct {
 		args    args
@@ -277,7 +275,6 @@ func TestHandleCheckCoupling(t *testing.T) {
 					"output_mode": "summary",
 				},
 			},
-			isError: &errFalse,
 		},
 		"happy_path_full": {
 			args: args{
@@ -288,7 +285,6 @@ func TestHandleCheckCoupling(t *testing.T) {
 					"output_mode": "full",
 				},
 			},
-			isError: &errFalse,
 		},
 		"invalid_arguments": {
 			args: args{
@@ -426,7 +422,6 @@ func TestHandleDetectClones(t *testing.T) {
 func TestHandleFindDeadCode(t *testing.T) {
 
 	errTrue := true
-	errFalse := false
 
 	tests := map[string]struct {
 		args    args
@@ -441,7 +436,6 @@ func TestHandleFindDeadCode(t *testing.T) {
 					return setupTestFile(t, "classes.py")
 				},
 			},
-			isError: &errFalse,
 		},
 		"happy_path_info": {
 			args: args{
@@ -452,7 +446,6 @@ func TestHandleFindDeadCode(t *testing.T) {
 					return setupTestFile(t, "classes.py")
 				},
 			},
-			isError: &errFalse,
 		},
 		"happy_path_critical": {
 			args: args{
@@ -463,7 +456,6 @@ func TestHandleFindDeadCode(t *testing.T) {
 					return setupTestFile(t, "classes.py")
 				},
 			},
-			isError: &errFalse,
 		},
 		"happy_path_default_severity": {
 			args: args{
@@ -472,7 +464,6 @@ func TestHandleFindDeadCode(t *testing.T) {
 				},
 				arguments: map[string]interface{}{},
 			},
-			isError: &errFalse,
 		},
 		"invalid_arguments": {
 			args: args{
@@ -526,7 +517,6 @@ func TestHandleFindDeadCode(t *testing.T) {
 func TestHandleGetHealthScore(t *testing.T) {
 
 	errTrue := true
-	errFalse := false
 
 	tests := map[string]struct {
 		args    args
@@ -539,7 +529,6 @@ func TestHandleGetHealthScore(t *testing.T) {
 				},
 				arguments: map[string]interface{}{},
 			},
-			isError: &errFalse,
 		},
 		"invalid_arguments": {
 			args: args{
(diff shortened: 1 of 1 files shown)


mcp/handlers_test.go
```diff
@@ -0,0 +1,667 @@
...
+					"output_mode": "full",
+				},
+			},
+			isError: &errFalse,
+		},
+		"invalid_arguments": {
+			args: args{
+				arguments: nil,
+			},
+			isError: &errTrue,
```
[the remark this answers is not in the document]
DaisukeYoda: Same issue in TestHandleGetHealthScore at L596-600, where "bad" is similarly converted to an empty map at L630-635.

mcp/dependencies.go
```diff
@@ -41,3 +41,11 @@ func (d *Dependencies) ConfigPath() string {
...
 	return buildAnalyzeUseCase(d.fileReader)
 }
+
+func NewTestDependencies(fr domain.FileReader, cfg *config.Config, path string) *Dependencies {
+	return &Dependencies{
+		fileReader: fr,
+		config:     cfg,
+		configPath: path,
+	}
+}
```
DaisukeYoda: NewTestDependencies is only used in tests but lives in production code. Consider using Go's https://go.dev/src/fmt/export_test.go to keep it out of the final binary.

mcp/handlers_test.go
DaisukeYoda: Suggestion: The setup logic (creating deps, building the request, injecting path, checking IsError) is nearly identical across all 6 test functions. Extracting a shared helper would cut down on duplication and make future changes easier.

Also, TestHandleCheckCoupling creates the test file eagerly at L357 rather than using the setupFS callback like the other five. Might be worth unifying for consistency.

mcp/dependencies.go
DaisukeYoda: For this case, moving NewTestDependencies to a _test.go file in package mcp is probably the simplest option

DaisukeYoda:
@BenjaminGoehry 

In `TestHandleCheckCoupling`, `TestHandleFindDeadCode`, and `TestHandleGetHealthScore`, the happy path cases set `isError: &errFalse`. Because of this, the test always takes the early return path and **the content assertions below are never executed**:

```go
if tc.isError != nil {       // &errFalse is non-nil, so this is always true
    require.Equal(t, *tc.isError, res.IsError)
    return                   // ← exits here
}

// Everything below is dead code for these test cases
require.False(t, res.IsError)
require.NotEmpty(t, res.Content)
assert.Contains(t, out, "health_score")  // ← never runs
```

`TestHandleCheckComplexity` works correctly because its success case leaves `isError` unset (nil), allowing the test to fall through to the content checks. The same approach should be applied to the other three test function.  You can just remove `isError: &errFalse` from the happy path cases.

DaisukeYoda:
Thank you very much!

PR #333, opened 2026-02-11
PR: Integrate TF-IDF + Cosine Similarity for Precision Type-2 Clone Detection

### LINKED ISSUE #315: Introduce TF-IDF + Cosine Similarity for Type-2 Clone Detection to Reduce False Positives
## Summary

### Benefits of TF-IDF

### Implementation Proposal

```go
type TFIDFCalculator struct {
}

    if df == 0 {
        return 0
    }
}

    tf := make(map[string]int)
    for _, f := range features {
        tf[f]++
    }
    
    for f, count := range tf {
    }
    return vector
}

// 3. Cosine similarity
    
    for k, val1 := range v1 {
        if val2, ok := v2[k]; ok {
            dotProduct += val1 * val2
        }
        norm1 += val1 * val1
    }
    for _, val2 := range v2 {
        norm2 += val2 * val2
    }
    
    if norm1 == 0 || norm2 == 0 {
        return 0
    }
}
```

### Expected Benefits

## Implementation Locations

## Compatibility

## Related

avitej-iyer:
Key changes : 

1. Architectural Refactoring 
    - SimilarityAnalyzer Interface: Created a new interface in internal/analyzer to decouple CloneDetector from specific mathematical strategies.
    - Switchable Strategies : Refactored NewCloneDetector to support selecting between the structural APTEDAnalyzer (default) and the new SyntacticSimilarityAnalyzer.

2. Core Logic Implementation
    - SyntacticSimilarityAnalyzer: A high-performance analyzer that uses AST feature extraction. It supports traditional Jaccard similarity and the new TF-IDF mode.
    - TFIDFCalculator: New utility to handle IDF pre-computation and convert feature sets into weighted vectors for Cosine Similarity.
    - CloneDetector Integration: Added a pre-computation phase to DetectClones to calculate global inverse document frequencies before performing pairwise comparisons.

3. Verification 
    - Created tfidf_integration_test.go
    - Verified that common Type-1/2/3 tests still pass under the default configuration, ensuring backward compatibility.

The feature is opt-in for this version. To enable it, set the UseTFIDF flag in your CloneDetectorConfig : 

``` go
config := &analyzer.CloneDetectorConfig{
    UseTFIDF: true,
    // ... rest of config
}
detector := analyzer.NewCloneDetector(config)````

### DIFF
--- a/internal/analyzer/clone_detector.go
+++ b/internal/analyzer/clone_detector.go
@@ -201,6 +201,8 @@ type CloneDetectorConfig struct {
 	// Framework pattern handling (reduces false positives for dataclass, Pydantic, etc.)
 	ReduceBoilerplateSimilarity bool    // Apply lower weight to boilerplate nodes (default: true)
 	BoilerplateMultiplier       float64 // Cost multiplier for boilerplate nodes (default: 0.1)
+
+	UseTFIDF 	bool
 }
 
 // DefaultCloneDetectorConfig returns default configuration
@@ -258,6 +260,8 @@ type CloneDetector struct {
 	fragments   []*CodeFragment
 	clonePairs  []*ClonePair
 	cloneGroups []*CloneGroup
+
+	tfidfCalculator *TFIDFCalculator
 }
 
 // NewCloneDetector creates a new clone detector with the given configuration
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

File: internal/analyzer/similarity_analyzer.go [hunk 1]
```diff
@@ -4,7 +4,10 @@ package analyzer
 // Each clone type should have its own analyzer implementation.
 type SimilarityAnalyzer interface {
 	// ComputeSimilarity returns a similarity score between 0.0 and 1.0
-	ComputeSimilarity(fragment1, fragment2 *CodeFragment) float64
+	ComputeSimilarity(f1, f2 *CodeFragment, calc *TFIDFCalculator) float64
```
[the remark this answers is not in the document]
DaisukeYoda: Instead, store the calculator inside SyntacticSimilarityAnalyzer:

```go
type SyntacticSimilarityAnalyzer struct {
    extractor *ASTFeatureExtractor
    converter *TreeConverter
    tfidfCalc *TFIDFCalculator // add here
}
```

This eliminates the interface change, the apted.go renames, apted_interface_impl.go, and ~30 nil arguments throughout the codebase.

internal/analyzer/syntactic_similarity.go
```diff
@@ -84,7 +86,7 @@ func (s *SyntacticSimilarityAnalyzer) ComputeDistance(f1, f2 *CodeFragment) floa
 	if f1 == nil || f2 == nil {
 		return 0.0
 	}
-	return 1.0 - s.ComputeSimilarity(f1, f2)
+	return 1.0 - s.ComputeSimilarity(f1, f2, nil)
```
DaisukeYoda: ComputeDistance always passes nil, so it uses Jaccard even in TF-IDF mode. This means 1 - Similarity != Distance in the same ClonePair.

internal/analyzer/clone_detector.go
```diff
@@ -549,15 +559,17 @@ func (cd *CloneDetector) DetectClonesWithContext(ctx context.Context, fragments
 	// Convert AST fragments to tree nodes
 	cd.prepareFragments()
 
-	// Check for cancellation after preparation
-	if isCancelled(ctx) {
-		return cd.clonePairs, cd.cloneGroups
```
DaisukeYoda: The cancellation check between prepareFragments() and detectClonePairsWithContext() was deleted. Please restore it.

DaisukeYoda (changes requested):
Thanks for the contribution! The core TF-IDF logic looks solid and addresses #315 well. Please check a few things to fix before merge.

DaisukeYoda:
It’s been a while since the last activity, so I’m closing this PR for now. If your college work has settled down, contributions are always welcome anytime!

PR #382, opened 2026-04-09
honor complexity config settings in analyze

### LINKED ISSUE #368: TOML config silently ignores `enabled` and `report_unchanged` in [complexity]
## Summary

The `[complexity]` section in `.pyscn.toml` (and `pyproject.toml [tool.pyscn]`) defines `enabled` and `report_unchanged` keys, but these values are **silently ignored** when loading config because the TOML parsing struct is missing the corresponding fields.

## Details

The TOML template (`internal/config/default_config.toml.tmpl`) generates:

```toml
[complexity]
enabled = true
report_unchanged = true
low_threshold = 10
medium_threshold = 20
```

The internal config struct (`internal/config/config.go:77-80`) correctly defines both fields:

```go
Enabled         bool `mapstructure:"enabled"`
ReportUnchanged bool `mapstructure:"report_unchanged"`
```

But the **TOML loading struct** (`internal/config/toml_loader.go:28-33`) is missing them:

```go
type ComplexityTomlConfig struct {
    LowThreshold    *int `toml:"low_threshold"`
    MediumThreshold *int `toml:"medium_threshold"`
    MaxComplexity   *int `toml:"max_complexity"`
    MinComplexity   *int `toml:"min_complexity"`
    // ← Enabled and ReportUnchanged are missing
}
```

This means setting `enabled = false` or `report_unchanged = false` in `.pyscn.toml` has **no effect**. The values fall back to hardcoded defaults.

Note: Other sections like `[dead_code]` correctly include `Enabled *bool` in their TOML struct — this is only missing for `[complexity]`.

## What to change

1. **`internal/config/toml_loader.go`** — Add the missing fields to `ComplexityTomlConfig`:
   ```go
   type ComplexityTomlConfig struct {
       Enabled         *bool `toml:"enabled"`
       ReportUnchanged *bool `toml:"report_unchanged"`
       LowThreshold    *int  `toml:"low_threshold"`
       MediumThreshold *int  `toml:"medium_threshold"`
       MaxComplexity   *int  `toml:"max_complexity"`
       MinComplexity   *int  `toml:"min_complexity"`
   }
   ```

2. **`internal/config/pyscn_config.go`** — Ensure `PyscnConfigToConfig()` maps these new fields to the internal config (check around the complexity conversion section)

3. **Add tests** — Verify that setting `enabled = false` and `report_unchanged = false` in a TOML config file actually takes effect

## How to verify

```bash
make test
```

Also manually verify by creating a `.pyscn.toml` with `enabled = false` under `[complexity]` and confirming that `pyscn analyze` skips complexity analysis.

M-Hassan-Raza:
Closes #368

This fixes the gap where the complexity config was getting loaded but not actually respected all the way through analyze.

[complexity].enabled and report_unchanged now flow through config loading, request conversion, and runtime behavior. I also fixed min_complexity precedence so the value from [complexity] doesn't get quietly overwritten by the default output filter, while an explicit [output].min_complexity still wins if that's what someone set.

I cleaned up the analyze/complexity handoff too so it isn't resolving the same config in a couple diff places.

Tests run:
go test ./internal/config
go test ./app -run 'TestAnalyzeUseCase_|TestComplexityUseCase_'
go test ./service -run 'TestConfigurationLoader_|TestComplexityService_'
go test ./...

### DIFF
--- a/app/analyze_usecase.go
+++ b/app/analyze_usecase.go
@@ -51,6 +51,19 @@ type AnalyzeUseCase struct {
 	errorCategorizer domain.ErrorCategorizer
 }
 
+type analyzeExecutionConfig struct {
+	includePatterns           []string
+	excludePatterns           []string
+	recursive                 bool
+	complexityEnabled         bool
+	reportUnchanged           bool
+	complexityLowThreshold    int
+	complexityMediumThreshold int
+	complexityMaxComplexity   int
+	lshEnabled                string
+	lshThreshold              int
+}
+
 // AnalyzeUseCaseBuilder builds an AnalyzeUseCase
 type AnalyzeUseCaseBuilder struct {
 	complexityUseCase *ComplexityUseCase
@@ -184,7 +197,7 @@ type AnalysisTask struct {
 func (uc *AnalyzeUseCase) Execute(ctx context.Context, useCaseCfg AnalyzeUseCaseConfig, paths []string) (*domain.AnalyzeResponse, error) {
 	startTime := time.Now()
 
-	// Resolve config path once so all analysis phases read the same file.
+	// Resolve config path once so file discovery and task setup use a single config source.
 	targetPath := ""
 	if len(paths) > 0 {
 		targetPath = paths[0]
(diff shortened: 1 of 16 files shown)


DaisukeYoda (approved):
LGTM — thanks for another great PR, really appreciate all the contributions lately!

PR #383, opened 2026-04-09
refactor: remove internal/config import from AnalyzeUseCase

### LINKED ISSUE #372: refactor: remove internal/config import from app/analyze_usecase.go
## Summary

`app/analyze_usecase.go` directly imports the `internal/config` package, violating the Clean Architecture dependency rule (app layer should only depend on domain).

## Current State

```go
// app/analyze_usecase.go
import (
    "github.com/ludo-technologies/pyscn/internal/config"
)
```

- Directly calls `config.NewTomlConfigLoader()` and `config.LoadConfig()`
- Other use cases (`ComplexityUseCase`, `DeadCodeUseCase`, etc.) correctly load configuration through domain interfaces — this violation is specific to `AnalyzeUseCase`

## Expected Fix

1. Remove the `internal/config` import from `analyze_usecase.go`
2. Replace direct config loading with domain interfaces (e.g., existing `ConfigurationLoader`)
3. If needed, define a new interface in the domain layer and implement it in the service layer

## References

- Correct pattern: `app/complexity_usecase.go` (uses `domain.ConfigurationLoader`)
- Correct pattern: `app/dead_code_usecase.go` (uses `domain.DeadCodeConfigurationLoader`)

M-Hassan-Raza:
Closes #372

This moves the analyze-specific config resolution/loading behind a domain interface so AnalyzeUseCase no longer reaches into internal/config directly.

I added an AnalyzeConfigurationLoader in the domain layer, implemented it in service, and wired AnalyzeUseCase to use that loader for config discovery and execution settings. The behavior stays the same, including resolved config path handling and the analyze defaults around .pyi files and clone LSH settings.

Tests run:
go test ./service -run 'TestAnalyzeConfigurationLoader_|TestConfigurationLoader_|TestComplexityService_'
go test ./app -run 'TestAnalyzeUseCase_|TestComplexityUseCase_'
go test ./cmd/pyscn -run 'TestAnalyze'
go test ./...

### DIFF
--- a/app/analyze_usecase.go
+++ b/app/analyze_usecase.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	"github.com/ludo-technologies/pyscn/domain"
-	"github.com/ludo-technologies/pyscn/internal/config"
 	"github.com/ludo-technologies/pyscn/service"
 )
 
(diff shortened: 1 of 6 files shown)


app/analyze_usecase.go
```diff
@@ -657,73 +639,29 @@ func (uc *AnalyzeUseCase) calculateSummary(summary *domain.AnalyzeSummary, respo
 	}
 }
 
-func (uc *AnalyzeUseCase) loadExecutionConfig(configPath string) (analyzeExecutionConfig, error) {
-	if configPath == "" {
-		return defaultAnalyzeExecutionConfig(), nil
+func (uc *AnalyzeUseCase) loadExecutionConfig(configPath string, paths []string) (domain.AnalyzeExecutionConfig, error) {
+	if uc.configLoader == nil {
```
DaisukeYoda: The two nil checks in loadExecutionConfig are both unreachable in practice. uc.configLoader is always set by Build(), so the guard here is dead code. If you want to catch a missing loader, make it a required-field error in Build() like fileReader. The executionCfg == nil check can be eliminated entirely by changing the loader's return type from *AnalyzeExecutionConfig to AnalyzeExecutionConfig; no implementation returns nil without an error.

DaisukeYoda:
@M-Hassan-Raza Hello, the implementation looks good. Could you look into my small suggestion?

M-Hassan-Raza:
> @M-Hassan-Raza Hello, the implementation looks good. Could you look into my small suggestion?

Should be resolved now, can you please chekc?

DaisukeYoda (approved):

DaisukeYoda:
Many thanks!!

PR #384, opened 2026-04-11
fix: honor clone show_content config in analyze

### LINKED ISSUE #132: [BUG] show_content configuration option is not implemented
**Describe the bug**

This option is documented and parsed, but it doesn´t seem to be used in the codebase ?

**To Reproduce**
1. Setup the config file with `show_content` option to `true` (in the output section)
2. Run command 'uvx pyscn analyze .' on a file with code duplication
3. GO in the html report, code duplication section
 
**Expected behavior**

In the HTML output, code duplication section, the duplicated code should be displayed

**Output**

Image

**Environment:**
 - OS: Ubuntu
 - pyscn Version: 1.0.0

**Additional context**

Thanks for the project !

It's a minor issue, but the main point is that it seems there's a mismatch between what the doc says in https://github.com/ludo-technologies/pyscn/blob/main/docs/ARCHITECTURE.md, what the default config file adds, and what the code actually does.

M-Hassan-Raza:
Closes #132

This fixes the clone config precedence so output settings from config don't get overwritten by request defaults.

The main problem was that clone config merging was happening in the use case with different behavior than the other analyzers, so `show_content` could be parsed from config but still never really take effect. I moved that merge logic into the clone config loaders, wired analyze through the same path, and kept the clone entry points consistent.

I also updated the analyze HTML clone section to actually render code previews when `show_content = true`, and made sure clone pair/group responses carry that content only when it's needed.

Tests I ran:
go test ./service -run 'TestCloneConfigurationLoader_|TestCloneConfigurationLoaderWithFlags_|TestCloneService_convertClone|TestAnalyzeFormatter_WriteHTML_|TestAnalyzeFormatter_Write_HTML'
go test ./app -run 'TestCloneUseCase_'
go test ./cmd/pyscn -run 'TestAnalyzeCommand'
go test ./domain ./service ./app ./cmd/pyscn
go test -p 1 ./...


DaisukeYoda (changes requested):
Thanks for the PR! The config centralization and HTML preview look great. I found one issue though.

The new `MergeConfig` in `CloneConfigurationLoader` handles more fields than the old `mergeConfiguration` did (e.g. `MaxSimilarity`, `GroupMode`, `GroupThreshold`). The old code simply didn't touch these fields, so they were preserved from the config. The new code compares them against `DefaultCloneRequest()` and treats any difference as "explicitly set by the user."

The problem is that `analyze_usecase.go` builds the clone request without setting these fields, so they end up as Go zero values (`0.0`, `""`). Since `0.0 != 1.0` (the default for `MaxSimilarity`), `MergeConfig` interprets that as an intentional override and writes `0.0` into the merged result. This causes `filterClonePairs` to drop every pair (`pair.Similarity > 0.0` is always true), so analyze reports zero clones.

I confirmed this by building both branches and running `pyscn analyze` on the same set of duplicate files. Main finds 15 clone pairs; this branch finds 0. The JSON output shows `max_similarity: 1` on main vs `max_similarity: 0` on this branch, with `group_mode` and `group_threshold` similarly zeroed out.

The simplest fix would be to have `analyze_usecase.go` initialize the clone request from `DefaultCloneRequest()` and only override what it actually needs, so unset fields match the defaults and pass through `MergeConfig` safely. Alternatively, `MergeConfig` could skip zero-value floats/strings when the base already has a non-zero value, but that feels more fragile.

Also, the `EnableDFA: true` addition in `DefaultCloneRequest()` seems unrelated to the config precedence fix — could you split that into a separate PR?

M-Hassan-Raza:
> Thanks for the PR! The config centralization and HTML preview look great. I found one issue though.
> 
> The new `MergeConfig` in `CloneConfigurationLoader` handles more fields than the old `mergeConfiguration` did (e.g. `MaxSimilarity`, `GroupMode`, `GroupThreshold`). The old code simply didn't touch these fields, so they were preserved from the config. The new code compares them against `DefaultCloneRequest()` and treats any difference as "explicitly set by the user."
> 
> The problem is that `analyze_usecase.go` builds the clone request without setting these fields, so they end up as Go zero values (`0.0`, `""`). Since `0.0 != 1.0` (the default for `MaxSimilarity`), `MergeConfig` interprets that as an intentional override and writes `0.0` into the merged result. This causes `filterClonePairs` to drop every pair (`pair.Similarity > 0.0` is always true), so analyze reports zero clones.
> 
> I confirmed this by building both branches and running `pyscn analyze` on the same set of duplicate files. Main finds 15 clone pairs; this branch finds 0. The JSON output shows `max_similarity: 1` on main vs `max_similarity: 0` on this branch, with `group_mode` and `group_threshold` similarly zeroed out.
> 
> The simplest fix would be to have `analyze_usecase.go` initialize the clone request from `DefaultCloneRequest()` and only override what it actually needs, so unset fields match the defaults and pass through `MergeConfig` safely. Alternatively, `MergeConfig` could skip zero-value floats/strings when the base already has a non-zero value, but that feels more fragile.
> 
> Also, the `EnableDFA: true` addition in `DefaultCloneRequest()` seems unrelated to the config precedence fix — could you split that into a separate PR?

Love the in depth feedback! Should be good now hopefuly. Might make another pr for that DFA logic bit

DaisukeYoda (approved):
LGTM!

PR #385, opened 2026-04-13
refactor: reuse visited state in reachability

### LINKED ISSUE #285: refactor(reachability): optimize markSuccessorsUnreachableAfterReturn with memoization
## Summary

`markSuccessorsUnreachableAfterReturn` function still uses `copyVisited` pattern which could lead to exponential time complexity in certain CFG structures.

## Location

`internal/analyzer/reachability.go:301`

```go
ra.markSuccessorsUnreachableAfterReturn(edge.To, result, copyVisited(visited))
```

## Context

This was discovered while fixing #282. The main issue was in `allSuccessorsReturn`, which has been fixed with memoization in #283.

`markSuccessorsUnreachableAfterReturn` has lower impact because:
- It's only called for blocks in `allPathsReturnBlocks` (blocks containing return statements)
- Not called for every block in the CFG

However, for codebases with many return statements and complex control flow, this could still cause performance issues.

## Suggested Fix

Apply the same pattern used in `findMaxDepthFrom` (coupling_metrics.go):
- Remove `copyVisited` 
- Add `defer func() { visited[block.ID] = false }()`

## Priority

P3 - Low priority, limited impact

M-Hassan-Raza:
Closes #285

This reuses the visited state in `markSuccessorsUnreachableAfterReturn` instead of copying it on every recursive step, which brings it in line with the existing backtracking pattern already used in the reachability analyzer.

I also added targeted coverage for the after-return path, including shared dead tails and a cyclic dead tail, plus a benchmark for the return-heavy walk so the change is easier to validate.

Ran:
go test ./internal/analyzer -run 'TestReachabilityAfterReturn|TestReachabilityEdgeCases|TestReachabilityWithRealCFG'
go test ./internal/analyzer -race -run 'TestReachabilityAfterReturn'
go test -p 1 ./...

### DIFF
--- a/internal/analyzer/reachability.go
+++ b/internal/analyzer/reachability.go
@@ -319,12 +319,3 @@ func (ra *ReachabilityAnalyzer) blockContainsReturn(block *BasicBlock) bool {
 
 	return false
 }
-
-// copyVisited creates a copy of the visited map to avoid sharing state between recursive calls
-func copyVisited(visited map[string]bool) map[string]bool {
-	copy := make(map[string]bool)
-	for k, v := range visited {
-		copy[k] = v
-	}
-	return copy
-}
--- a/internal/analyzer/reachability.go
+++ b/internal/analyzer/reachability.go
@@ -288,6 +288,7 @@ func (ra *ReachabilityAnalyzer) markSuccessorsUnreachableAfterReturn(block *Basi
 		return
 	}
 	visited[block.ID] = true
+	defer func() { visited[block.ID] = false }()
 
 	// If this block contains a return, its fall-through successors are unreachable
 	if ra.blockContainsReturn(block) {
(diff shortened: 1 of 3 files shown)


internal/analyzer/reachability.go
```diff
@@ -288,6 +288,7 @@ func (ra *ReachabilityAnalyzer) markSuccessorsUnreachableAfterReturn(block *Basi
 		return
 	}
 	visited[block.ID] = true
+	defer func() { visited[block.ID] = false }()
```
DaisukeYoda: Backtracking (`defer { visited[block.ID] = false }`) is correct in `allSuccessorsReturn`, which is a pure predicate that must explore each path independently. Here, though, the function is a side-effecting traversal (deleting from `ReachableBlocks`). Unwinding visited means diamond-shaped graphs cause the same block to be processed multiple times, which can degrade to exponential visits.

Since the only purpose of `visited` here is cycle detection, the simplest fix is to just remove this `defer` line entirely. `visited[block.ID] = true` on the line above already prevents revisits, and each block gets processed exactly once:

 ```go
 visited[block.ID] = true
 // no backtracking needed: this is a mark-and-sweep traversal, not a path predicate
 ```

Tests should still pass as-is since `delete()` is idempotent, but this avoids redundant work and prevents future breakage if non-idempotent logic is added.

DaisukeYoda (changes requested):
Thanks for the steady stream of PRs! Please take a look at my review.

M-Hassan-Raza:
should be good to go now!

DaisukeYoda (approved):
LGTM! Thanks!

PR #389, opened 2026-04-16
fix: make grouping clone-type tie breaks deterministic

M-Hassan-Raza:
This is a small follow-up I found while validating the complete-linkage work.

When multiple clone types had the same vote count inside a group, `majorityCloneType` could return different results depending on Go map iteration order. That made the output nondeterministic in tied cases.

This change makes the tie-break explicit and stable by preferring the lower `CloneType` when counts are equal.

Tests:
- add coverage for tied clone-type votes
- verify the result is deterministic in the tie case

Back with another PR after a short hiatus, this time for a small determinism cleanup before the larger grouping work.

### DIFF
--- a/internal/analyzer/connected_grouping.go
+++ b/internal/analyzer/connected_grouping.go
@@ -152,7 +152,7 @@ func majorityCloneType(typeMap map[string]CloneType, members []*CodeFragment) Cl
 	var best CloneType
 	maxC := -1
 	for t, c := range counts {
-		if c > maxC || (c == maxC && (best == 0 || t < best)) {
+		if c > maxC || (c == maxC && t < best) {
 			maxC = c
 			best = t
 		}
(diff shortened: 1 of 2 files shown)


DaisukeYoda (changes requested):
Small nit: the `best == 0` guard in the tie-break is unreachable. The first iteration always enters `c > maxC` (counts are >= 1 and maxC starts at -1), so `best` is set before any tie can occur. Consider simplifying to `c == maxC && t < best`.

Thanks!

M-Hassan-Raza:
> Small nit: the `best == 0` guard in the tie-break is unreachable. The first iteration always enters `c > maxC` (counts are >= 1 and maxC starts at -1), so `best` is set before any tie can occur. Consider simplifying to `c == maxC && t < best`.
> 
> Thanks!

Should be good now, thanks!

DaisukeYoda (approved):

DaisukeYoda:
Many thanks!

Issue #86, opened 2025-09-11
Optimize Complete Linkage clustering with Lance-Williams update formula


DaisukeYoda:
## Problem

The current Complete Linkage grouping implementation in `internal/analyzer/complete_linkage_grouping.go` recalculates all cluster-pair similarities from scratch after each merge operation, leading to O(n³m²) time complexity for large datasets.

## Current Implementation Issues

```go
// Current: Recalculates all similarities every iteration
for {
    for i := 0; i < len(clusters); i++ {
        for j := i + 1; j < len(clusters); j++ {
            s := clusterSim(clusters[i], clusters[j])  // O(m²) every time
        }
    }
    // Merge best pair and repeat...
}
```

**Time Complexity:** O(n³m²) where n = number of clusters, m = average cluster size

## Proposed Solution: Lance-Williams Update Formula

Implement the Lance-Williams distance update formula for Complete Linkage:

```
dist(AB, C) = max(dist(A, C), dist(B, C))
```

Where AB is the merged cluster of A and B.

### Benefits:
- **Reduced complexity:** O(n²m²) (removes one factor of n)
- **Memory efficiency:** Reuse existing distance calculations
- **Standard approach:** Well-established algorithm in hierarchical clustering

### Implementation Plan:
1. Cache inter-cluster similarities in a distance matrix
2. After each merge, update distances using Lance-Williams formula instead of recalculating
3. Maintain backward compatibility with existing GroupingStrategy interface

## Performance Impact

**Before:** 1000 fragments → ~25 billion similarity calculations
**After:** 1000 fragments → ~25 million similarity calculations (1000x improvement)

## References
- Lance, G.N. & Williams, W.T. (1967). A general theory of classificatory sorting strategies
- Standard hierarchical clustering optimization used in scipy, R, etc.

**Priority:** Medium (performance optimization for large datasets)
**Difficulty:** Medium (algorithmic change, needs careful testing)

M-Hassan-Raza:
I think this can be closed safely. What do you think @DaisukeYoda?

DaisukeYoda:
@M-Hassan-Raza Yes, thanks!

PR #391, opened 2026-04-18
perf: optimize complete-linkage grouping

M-Hassan-Raza:
This finishes the deeper part of #86.

Complete-linkage grouping now keeps sparse threshold-qualified neighbors and uses an indexed best-neighbor heap instead of rescanning every active cluster pair after each merge.

On `BenchmarkGroupingStrategies/completelink_1000` (`main` -> this branch):

- `time`: `11,091,543,992 ns/op` -> `889,650 ns/op`
- `memory`: `5,093,308,049 B/op` -> `1,216,729 B/op`
- `allocs`: `228,070,871 allocs/op` -> `12,531 allocs/op`

I also tightened coverage around the new path with:

- parity checks against the slow reference implementation
- `4,096` exhaustive small-graph cases
- `72` randomized parity cases across `3` thresholds
- determinism checks across `32` runs
- clique and non-mergeable output invaraints
- dense benchmark coverage

This takes the complete-linkage fix all the way instead of only fixing the inner recompuation path.

### DIFF
--- a/internal/analyzer/complete_linkage_clusterer.go
+++ b/internal/analyzer/complete_linkage_clusterer.go
@@ -1,0 +1,170 @@
+package analyzer
+
+// completeLinkageClusterer stores only threshold-qualified inter-cluster edges.
+// That keeps sparse workloads sparse while still supporting exact complete-linkage
+// merges, since a merged cluster can stay adjacent to C only if both source
+// clusters already had qualifying edges to C.
+type completeLinkageClusterer struct {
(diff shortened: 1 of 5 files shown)


DaisukeYoda (approved):
LGTM! Huge speedup and the tests look solid.

M-Hassan-Raza:
> LGTM! Huge speedup and the tests look solid.

I'm glad it all came together nicely after a ton of iterations and testing 🙌

PR #401, opened 2026-04-29
fix: make clone trees expression-aware

### LINKED ISSUE #400: Clone detection: ConvertAST drops expression-level AST fields, leading to inflated structural similarity
## Summary

`internal/analyzer/apted_tree.go`'s `ConvertAST` only descends through `Children/Body/Orelse/Finalbody/Handlers`. The parser stores most expression-level content in *other* fields (`Test`, `Targets`, `Iter`, `Args`, `Keywords`, `Decorator`, `Bases`, `Left`, `Right`, `Value` when `*parser.Node`), and several pieces of identifier/literal information live as **plain strings** on the node (e.g. `Attribute.Name`, `Keyword.Name`, `Arg.Name`, `with as` aliases, `except as` aliases, import aliases).

As a result the APTED tree built for clone detection is essentially a **statement-type skeleton** — it does not see assignment targets, subscript indices, function arguments, attribute names, comparison operands, etc. Two fragments with the same sequence of statement types compare as APTED-distance 0 regardless of their actual contents.

This was the underlying mechanism behind the false `type:1, similarity:1.0, distance:0` reports investigated in #395. PR #399 hides the most user-visible symptom by requiring textual identity for Type-1, but the structural similarity score remains over-confident and `distance` is still reported as `0` for many unrelated pairs.

## Reproduction

`centrosome/cpmorphology.py:1125-1161` vs `1169-1205` — two parallel `if`-blocks with `i`↔`j` identifiers swapped throughout the body. Even with PR #399 the pair is reported with `distance: 0` because the APTED tree is identical for both blocks (no `Name(...)` nodes ever enter the tree).

A self-contained Go-level reproduction:

```go
frag1 := `
if cond:
    a = b + c
    d = e[i]
    for n in range(k):
        a = a + d
`
frag2 := `
if cond:
    a = b + c
    d = e[j]    # only difference
    for n in range(k):
        a = a + d
`
// after parsing + ConvertAST: tree size = 7 nodes for both, all labels are
// {If, Assign, Assign, For, Assign}; no Name/Constant/BinOp/Subscript labels.
// APTED ComputeDistance returns 0.0, ComputeSimilarity returns 1.0.
```

## Findings collected during the #395 investigation

Initial fix: extending `ConvertAST` to descend into the missing pointer fields *does* let APTED see identifier and literal differences (distance becomes non-zero, e.g. 3.5 in the toy repro above, similarity ≈ 0.97). However:

1. **Scalar string payloads still missing.** `Attribute.Name`, `Keyword.Name`, `Arg.Name`, alias fields and `Module`/`Names` on imports are stored as `string`, not `*Node`, so a pointer-traversal fix alone leaves `obj.foo` vs `obj.bar` and `f(x=1)` vs `f(y=1)` indistinguishable. These need to be encoded in the node label or as synthetic leaf nodes.

2. **`calculateASTSize` (clone_detector.go:82) uses the old shallow traversal** (`Children/Body/Orelse` only). With richer trees, `MinNodes`/`shouldCompareFragments` no longer reflect actual fragment size — small but expression-heavy fragments slip past size filters.

3. **LSH/Jaccard candidate explosion.** Naively walking expression fields adds many `Name(*)`/`Constant(*)` leaves whose canonical labels collapse to universal tokens. MinHash signatures lose entropy, LSH buckets balloon, and `seenPairs` in `DetectClonesWithLSH` (clone_detector.go:675) tracks O(n²) candidates. In a centrosome run the naive fix took >30 minutes vs 52 seconds pre-fix despite per-pair APTED actually being faster (167ms→42ms benchmarked).

4. **Confidence/CloneType decoupling in `compareFragmentsWithClassifier`** (clone_detector.go ~944-968) — `result.Confidence` comes from the cascading classifier while `CloneType` is re-derived from APTED similarity. PR #399 makes the type derivation textual-aware but the dual sourcing of these fields is still latent.

## Suggested approach

A single PR that lands all of (1)–(4) at once is risky; the investigation showed the perf cliff is real. A staged plan:

1. Centralize child enumeration in one ordered helper (groundwork; no behavioral change).
2. Encode scalar payloads in node labels or as synthetic leaves so identifier/literal differences are visible to APTED.
3. Walk the missing pointer fields in `ConvertAST`.
4. Update `ASTFeatureExtractor` so the richer tree does not flood Jaccard/MinHash with universal tokens (e.g. drop `Name`/`Constant` leaves from feature emission, or weight by IDF).
5. Cap LSH bucket sizes / drop oversized buckets so a pathological signature distribution cannot fan out.
6. Update `calculateASTSize` and any other consumers that assume the old skeleton-only tree shape.

Each step should be measured against a representative corpus (centrosome is a useful benchmark; pre-fix baseline was 52s for `pyscn analyze .`).

## Acceptance

- Two fragments that share the same statement skeleton but differ in identifiers, literals, attribute names, keyword argument names, or subscript indices have **non-zero** APTED distance.
- `pyscn analyze` runtime on the centrosome corpus stays within ~2× of the pre-#395 baseline.
- Existing Type-2 detection tests continue to pass (renamed-identifier code is still recognized as Type-2).

## Context

- Discovered while investigating #395.
- PR #399 ships the user-visible Type-1 fix without depending on this redesign — keeping the two efforts independent is intentional.
- Independent code review (codex) corroborated findings (1)–(3); see investigation log for details.

M-Hassan-Raza:
Closes #400

Convert clone APTED trees with expression-level AST fields instead of only the statement skeleton.

This makes identifier, literal, attribute, keyword, operator, import, alias, annotation, decorator, and comprehension differences affect the reported distance.

It also keeps LSH bounded by filtering low-entropy feature leaves and capping per-query candidates without dropping large exact-clone families.

Validation: go test ./...; go vet ./...; expression edge matrix 33/33 non-zero vs upstream 0/33; LSH exact-family cap regression test.

### DIFF
--- a/internal/analyzer/apted_tree.go
+++ b/internal/analyzer/apted_tree.go
@@ -2,6 +2,8 @@ package analyzer
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/ludo-technologies/pyscn/internal/parser"
 )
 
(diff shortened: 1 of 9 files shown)


DaisukeYoda (approved):
Excellent work! When I filed #400 I expected the fix to come with a performance cliff — landing both in one PR is genuinely impressive.

Verified locally: the `i`↔`j` reproducer from #400 now reports `distance: 3.5` (was `0`), the pair correctly demotes to Type-2, and centrosome finishes in 162s (earlier exploratory attempts here didn't finish in 30 minutes). 

Thank you for picking this up. LGTM!

M-Hassan-Raza:
> Excellent work! When I filed #400 I expected the fix to come with a performance cliff — landing both in one PR is genuinely impressive.
> 
> Verified locally: the `i`↔`j` reproducer from #400 now reports `distance: 3.5` (was `0`), the pair correctly demotes to Type-2, and centrosome finishes in 162s (earlier exploratory attempts here didn't finish in 30 minutes).
> 
> Thank you for picking this up. LGTM!

I'm glad it came out pretty well 😁 It took quite a while to get it into shape and test the performance, would've hated to hurt the perf side of the tool. happy ti didn't come to that

PR #415, opened 2026-05-07
fix: preserve f-string ivar access

### LINKED ISSUE #413: [BUG][auto] lcom4-misses-self-access-in-f-string: LCOM4 ignores self.<ivar> accesses inside f-string interpolations
## Bug

LCOM4 misses `self.<ivar>` accesses that appear inside f-string interpolations (`f"{self._x}"`). Methods that access an instance variable only via an f-string are reported as not sharing the ivar with other methods that *do* access it, inflating LCOM4 by reporting falsely-disconnected method components.

This is the same family of bug as #412 (different AST node missed by the LCOM4 ivar-collection visitor). #412 covers `with`/`async with` context expressions; this issue covers `JoinedStr`/`FormattedValue` (f-string) AST nodes.

## Repro (minimal synthetic)

```python
# repro.py
class A:
    """ivar accessed in f-string only — should be LCOM4=1"""
    def __init__(self):
        self._sep = '/'
    def render(self):
        return f"x{self._sep}y"


class B:
    """ivar accessed in f-string format spec — should be LCOM4=1"""
    def __init__(self):
        self._sep = '/'
    def render(self):
        return f"x{1:{self._sep}>5}y"


class C:
    """control: plain string concatenation — correctly reports LCOM4=1"""
    def __init__(self):
        self._sep = '/'
    def render(self):
        return "x" + self._sep + "y"


class D:
    """mixed: m1 direct access (detected), m2 f-string (NOT detected)"""
    def __init__(self):
        self._x = 1
    def m1(self):
        return self._x
    def m2(self):
        return f"{self._x}"
```

```bash
$ ./pyscn analyze --json --select lcom repro.py
```

## Expected output

```json
{"Name": "A", "Metrics": {"LCOM4": 1, "MethodGroups": [["__init__","render"]]}}
{"Name": "B", "Metrics": {"LCOM4": 1, "MethodGroups": [["__init__","render"]]}}
{"Name": "C", "Metrics": {"LCOM4": 1, "MethodGroups": [["__init__","render"]]}}
{"Name": "D", "Metrics": {"LCOM4": 1, "MethodGroups": [["__init__","m1","m2"]]}}
```

## Actual output (pyscn v1.18.1-13-gae064f8)

```json
{"Name": "A", "Metrics": {"LCOM4": 2, "MethodGroups": [["__init__"],["render"]]}}      // ❌ should be 1
{"Name": "B", "Metrics": {"LCOM4": 2, "MethodGroups": [["__init__"],["render"]]}}      // ❌ should be 1
{"Name": "C", "Metrics": {"LCOM4": 1, "MethodGroups": [["__init__","render"]]}}        // ✅ correct
{"Name": "D", "Metrics": {"LCOM4": 2, "MethodGroups": [["__init__","m1"],["m2"]]}}    // ❌ should be 1
```

D is the smoking gun: `m1` reads `self._x` directly and gets connected to `__init__`. `m2` reads `self._x` only via `f"{self._x}"` — pyscn doesn't see that reference, so `m2` ends up in its own component.

## Bug surface

| Pattern | Sample | LCOM4 | Verdict |
|---|---|---|---|
| `return self._x` | direct return | 1 | ✅ |
| f-string | `f"{self._x}"` | **2** | ❌ |
| f-string with format spec | `f"{1:{self._sep}>5}"` | **2** | ❌ |
| f-string inside other call | `print(f"{self._x}")` | **2** | ❌ |
| plain concat | `"a" + self._x` | 1 | ✅ |

The pattern: `self.<ivar>` accesses that are descendants of a `JoinedStr` AST node are not visited.

## Related

This is a sibling of #412 (`lcom4-misses-self-access-in-with-statement`). Both bugs share the same root cause type — the LCOM4 ivar-collection AST visitor doesn't recurse into certain expression contexts. Distinct from #412 because the fix sites in code are likely different (different AST node types). Worth separate tracking so a fix for one doesn't silently miss the other.

Other contexts I tested that work correctly (visitor descends into them): `if` cond, `while` cond, function-call args, attribute-chain access, `for` cond, plain string concat. The two known broken contexts are: `With`/`AsyncWith` context expressions (#412) and `JoinedStr`/`FormattedValue` (this issue).

## How this was found

Found via the FP-audit skill in repo `tonybaloney/wily@2522acfe`. The class `src/wily/commands/build.py:SpeedColumn` (2 methods sharing `self.separator`, with `render` accessing it via `f"{...}{self.separator}{...}"`) is reported with LCOM4=2 against an actual cohesion of 1. Synthetic repro above isolates the trigger to f-string interpolation.

## Environment

- pyscn version: `v1.18.1-13-gae064f8`
- Platform: Darwin 24.3.0 (macOS)
- Python source under analysis: 3.x compatible

hussainarslan:
## What:
- Preserve f-string interpolation expressions in the parser AST
- Keep format specs, conversions, and debug markers visible to AST consumers
- Use shared deep traversal for LCOM instance-variable collection
- Add parser, LCOM, and APTED regression coverage

## Why:
LCOM4 missed self.<ivar> accesses inside f-strings because tree-sitter string nodes with interpolations were collapsed to constants before analyzers could visit their expressions.

Closes #413

### DIFF
--- a/internal/parser/ast_builder.go
+++ b/internal/parser/ast_builder.go
@@ -166,7 +166,12 @@ func (b *ASTBuilder) buildNode(tsNode *sitter.Node) *Node {
 		return b.buildName(tsNode)
 	case "integer", "float", "true", "false", "none":
 		return b.buildConstant(tsNode)
-	case "string", "concatenated_string":
+	case "string":
+		if b.stringHasInterpolation(tsNode) {
+			return b.buildFormattedString(tsNode)
+		}
+		return b.buildConstant(tsNode)
+	case "concatenated_string":
 		if b.hasStringInterpolation(tsNode) {
 			return b.buildFormattedString(tsNode)
 		}
@@ -1554,14 +1559,56 @@ func (b *ASTBuilder) hasStringInterpolation(tsNode *sitter.Node) bool {
 	if tsNode == nil {
 		return false
 	}
-	if tsNode.Type() == "interpolation" {
+	switch tsNode.Type() {
+	case "interpolation":
 		return true
+	case "string":
+		return b.stringHasInterpolation(tsNode)
 	}
 
 	childCount := int(tsNode.ChildCount())
 	for i := 0; i < childCount; i++ {
 		if b.hasStringInterpolation(tsNode.Child(i)) {
 			return true
+		}
+	}
+	return false
+}
+
+func (b *ASTBuilder) stringHasInterpolation(tsNode *sitter.Node) bool {
+	if !b.hasFormattedStringPrefix(tsNode) {
+		return false
+	}
+
+	childCount := int(tsNode.ChildCount())
+	for i := 0; i < childCount; i++ {
+		if b.hasStringInterpolation(tsNode.Child(i)) {
+			return true
+		}
+	}
+	return false
+}
+
+func (b *ASTBuilder) hasFormattedStringPrefix(tsNode *sitter.Node) bool {
+	start := int(tsNode.StartByte())
+	end := int(tsNode.EndByte())
+	if start < 0 || start >= len(b.source) {
+		return false
+	}
+	if end > len(b.source) {
+		end = len(b.source)
+	}
+
+	for i := start; i < end; i++ {
+		switch b.source[i] {
+		case 'f', 'F':
+			return true
+		case 'r', 'R', 'u', 'U', 'b', 'B':
+			continue
+		case '\'', '"':
+			return false
+		default:
+			return false
 		}
 	}
 	return false
--- a/internal/parser/ast_builder.go
+++ b/internal/parser/ast_builder.go
@@ -1480,12 +1480,12 @@ func (b *ASTBuilder) buildFormattedValue(tsNode *sitter.Node) *Node {
 		if expr == nil {
 			continue
 		}
+		cursor = child.EndByte()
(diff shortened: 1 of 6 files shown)


internal/parser/ast_builder.go
```diff
@@ -1399,35 +1411,225 @@ func (b *ASTBuilder) buildConstant(tsNode *sitter.Node) *Node {
...
+		expr := b.buildNode(child)
+		if expr == nil {
+			continue
+		}
+		if node.Value == nil {
+			node.Value = expr
+			continue
+		}
+		node.AddChild(expr)
+		cursor = child.EndByte()
```
DaisukeYoda: In this case, cursor must be updated regardless of which branch the if takes. If the continue path inside the if skips the cursor update, the bytes already consumed by the expression get counted again on the next iteration, producing a duplicate literal.

DaisukeYoda (changes requested):
Thanks for raising this PR! The implementation looks good overall, with one small issue. Please take a look at my review comment.

internal/parser/ast_builder.go
hussainarslan: @DaisukeYoda This should do the trick

DaisukeYoda (approved):
LGTM, and the new tests look good too. Thanks!

PR #425, opened 2026-05-12
fix: honor analyze config flags

M-Hassan-Raza:
Small cleanup for the unified analyze command.

This makes analyze respect enabled flags from config, validates --select values, and keeps skip flags working when --select is used.

Also keeps the default broad analyze behavior when a config file does not mention system/dependency settings.

Tested with go test ./...


DaisukeYoda (approved):
LGTM! Thanks!

PR #435, opened 2026-05-15
Fix Type-4 semantic clone detection

M-Hassan-Raza:
What changed
- Unified parser child traversal so typed fields like Value participate consistently in traversal and APTED sizing.
- Updated DFA extraction to read parser field contracts for parameters, calls, subscripts, unary/named expressions, and keyword args.
- Preserved semantic Type-4 classifier results in detector reporting and added fixture-backed positive/negative coverage.

Why
Type-4 semantic scores could pass the classifier, then get dropped when the detector reclassified the pair with structural APTED. A few parser/DFA paths also missed expression nodes stored outside raw Children.

Validation
- go test ./internal/parser ./internal/analyzer -count=1
- go test ./... -count=1

### DIFF
--- a/internal/analyzer/clone_detector.go
+++ b/internal/analyzer/clone_detector.go
@@ -901,6 +901,10 @@ func (cd *CloneDetector) compareFragments(fragment1, fragment2 *CodeFragment) *C
 		return nil
 	}
 
+	if cd.usesSemanticClassifier() {
+		return cd.compareFragmentsWithClassifier(fragment1, fragment2)
+	}
+
 	// Early filtering check
 	if !cd.shouldCompareFragments(fragment1, fragment2) {
 		return nil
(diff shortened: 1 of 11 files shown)


internal/parser/ast.go
```diff
@@ -169,38 +169,96 @@ func (n *Node) AddToBody(node *Node) {
...
+	appendNodes(node.Bases)
+	appendNodes(node.Args)
+	appendNodes(node.Targets)
+	appendNode(node.Test)
+	appendNode(node.Iter)
+	appendNode(node.Left)
+	appendNode(node.Right)
+	appendNode(node.Target)
+	if valueNode, ok := node.Value.(*Node); ok {
+		appendNode(valueNode)
```
DaisukeYoda, line 219: This PR fixes the shared GetChildren() function in order to fix Type-4 detection, but since that shared function is used by other features as well, the behavior of those other features ends up changing as a side effect. I verified this locally — this is a change that significantly exceeds the  scope of the PR, but since it counts as an accuracy improvement, it's OK!

internal/analyzer/clone_detector.go
```diff
@@ -901,6 +901,10 @@ func (cd *CloneDetector) compareFragments(fragment1, fragment2 *CodeFragment) *C
 		return nil
 	}
 
+	if cd.usesSemanticClassifier() {
+		return cd.compareFragmentsWithClassifier(fragment1, fragment2)
+	}
+
```
DaisukeYoda, line 907: This change should improve accuracy, but it has a significant performance impact (the early return here skips both shouldCompareFragments and the Jaccard pre-filter), so every pair now goes straight to APTED (O(N³) per pair). I'll approve for now, but we'll continue to verify on our side whether there are any performance concerns.

DaisukeYoda (approved):
@M-Hassan-Raza 

Thank you for creating this PR! This PR seems to have a fairly big impact on the algorithm's behavior, so I reviewed it carefully. In conclusion, I believe this PR definitely improves the accuracy of the analysis. LGTM!

M-Hassan-Raza:
> @M-Hassan-Raza
> 
> Thank you for creating this PR! This PR seems to have a fairly big impact on the algorithm's behavior, so I reviewed it carefully. In conclusion, I believe this PR definitely improves the accuracy of the analysis. LGTM!

Glad to hear it! And for your comment about performance, I am very particualr about performance as well. I'll do another round of smoke checks across the codebase and try to squeeze out more performance where possible :D

PR #437, opened 2026-05-16
fix: exit non-zero when output generation fails in analyze command (refs #249)

CatfishGG:
## Summary

When `generateOutput` fails (e.g., cannot create output directory or write the report file), the error was printed to stderr but swallowed — the CLI still exited 0. Now the output error is returned so the exit status correctly reflects failure.

## Changes

- **cmd/pyscn/analyze.go**: Track `outputErr` from `generateOutput`. If analysis succeeds but output generation fails, return the output error (non-zero exit).
- **cmd/pyscn/utils_test.go**: New unit tests for `generateOutputFilePath` and `resolveOutputDirectory` covering:
  - Default directory (CWD/.pyscn/reports) creation and path resolution
  - Filename format (`analyze_<timestamp>.html`)
  - Different command/extension combinations

## Root cause

`runAnalyze()` called `generateOutput()` which could fail, but only printed a warning and returned `nil` → exit 0. Users in non-interactive environments (WSL, CI) never saw the warning.

Refs #249

### DIFF
--- a/cmd/pyscn/analyze.go
+++ b/cmd/pyscn/analyze.go
@@ -170,10 +170,12 @@ func (c *AnalyzeCommand) runAnalyze(cmd *cobra.Command, args []string) error {
 	response, analysisErr := useCase.Execute(ctx, config, args)
 
 	// Generate output even if there were partial failures
+	var outputErr error
 	if response != nil {
 		// Generate output
 		if err := c.generateOutput(cmd, response, args); err != nil {
 			fmt.Fprintf(cmd.ErrOrStderr(), "Warning: Failed to generate output: %v\n", err)
+			outputErr = err
 		}
 
 		// Print summary
@@ -183,6 +185,11 @@ func (c *AnalyzeCommand) runAnalyze(cmd *cobra.Command, args []string) error {
 	// Return the analysis error so CLI exits with non-zero status
 	if analysisErr != nil {
 		return analysisErr
+	}
+
+	// Return output error if analysis succeeded but output generation failed
+	if outputErr != nil {
+		return outputErr
 	}
 
 	return nil
--- a/cmd/pyscn/analyze.go
+++ b/cmd/pyscn/analyze.go
@@ -174,7 +174,6 @@ func (c *AnalyzeCommand) runAnalyze(cmd *cobra.Command, args []string) error {
 	if response != nil {
 		// Generate output
 		if err := c.generateOutput(cmd, response, args); err != nil {
-			fmt.Fprintf(cmd.ErrOrStderr(), "Warning: Failed to generate output: %v\n", err)
 			outputErr = err
 		}
 
--- a/cmd/pyscn/utils_test.go
+++ b/cmd/pyscn/utils_test.go
@@ -7,8 +7,13 @@ import (
 )
 
 func TestGenerateOutputFilePath_CreatesDefaultDirectory(t *testing.T) {
-	// Create a temporary directory to work in
+	// macOS: t.TempDir() returns /var/folders/... but os.Getwd() after Chdir
+	// returns /private/var/folders/... — normalise with EvalSymlinks
 	tempDir := t.TempDir()
+	tempDir, err := filepath.EvalSymlinks(tempDir)
+	if err != nil {
+		t.Fatalf("failed to eval symlinks on tempDir: %v", err)
+	}
 
 	// Change to the temp directory
 	oldCwd, err := os.Getwd()
@@ -50,8 +55,13 @@ func TestGenerateOutputFilePath_CreatesDefaultDirectory(t *testing.T) {
 }
 
 func TestResolveOutputDirectory_DefaultToCWD(t *testing.T) {
-	// Create a temporary directory to work in
+	// macOS: t.TempDir() returns /var/folders/... but os.Getwd() after Chdir
+	// returns /private/var/folders/... — normalise with EvalSymlinks
 	tempDir := t.TempDir()
+	tempDir, err := filepath.EvalSymlinks(tempDir)
+	if err != nil {
+		t.Fatalf("failed to eval symlinks on tempDir: %v", err)
+	}
 
 	// Change to the temp directory
 	oldCwd, err := os.Getwd()
@@ -78,7 +88,13 @@ func TestResolveOutputDirectory_DefaultToCWD(t *testing.T) {
 }
 
 func TestGenerateOutputFilePath_DifferentExtensions(t *testing.T) {
+	// macOS: t.TempDir() returns /var/folders/... but os.Getwd() after Chdir
+	// returns /private/var/folders/... — normalise with EvalSymlinks
 	tempDir := t.TempDir()
+	tempDir, err := filepath.EvalSymlinks(tempDir)
+	if err != nil {
+		t.Fatalf("failed to eval symlinks on tempDir: %v", err)
+	}
 
 	oldCwd, err := os.Getwd()
 	if err != nil {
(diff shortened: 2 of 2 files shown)


CatfishGG:
Tested on Docker with Ubuntu to simulate a Linux environment. Running 'uvx pyscn analyze .' in a non-interactive shell produced the HTML report correctly at '.pyscn/reports/analyze_<timestamp>.html' with exit code 0. The fix for error swallowing (return non-zero when output generation fails) is still valid and useful regardless, since it ensures CI pipelines and scripts can detect output failures.

However, the exact WSL/DrvFs interaction cannot be replicated in Docker. Possible WSL-specific causes:
1. DrvFs filesystem permissions on /mnt/c/Users/ paths
2. os.Getwd() returning Windows-style paths that Python on WSL struggles with
3. Symlinks or junction points silently causing os.MkdirAll to fail

The original bug was reported Dec 7 2025, but the default directory creation fix (commit db15d3c, Sep 12 2025) is already included in v1.20.0. If the issue persists on actual WSL, further investigation into the filesystem path resolution would be needed.

aftab1337:
Hmm, interesting.. Looks like it needs to be dug up further,

aftab1337:
@DaisukeYoda @M-Hassan-Raza what are your thoughts on this?

M-Hassan-Raza:
I’m mostly aligned with @aftab1337 here. This change is useful, because swallowing a report write failure and still exiting 0 is definitely wrong.

But I’m not fully convicned it closes #249 yet. The original report was WSL + uvx + no report + no obvious warning. This PR makes output failures fail harder, but it doesn’t really prove why the report was missing in that WSL case.

I’d probably frame this as “fix output errors being swallowed” and keep the WSL issue open until someone can repro on actual WSL. Would be useful to know if the project was under /mnt/c or /home, what `uvx pyscn --version` shows, the real exit code, and whether `.pyscn/reports` exists after running it.

A small PSA that i don't have extensive experience working with WSL so these are my general observations due to my limited understanding of that ecosystem.

CatfishGG:
Fair point. To be clear, this PR addresses the error-swallowing bug (output failures returning exit 0) which is a legitimate fix regardless of the original WSL issue.

However, you're right that this doesn't fully close #249 if the original problem was the report not being generated at all rather than just a bad exit code. My Docker testing confirmed that under normal Linux conditions the HTML generates correctly, so the WSL-specific failure mode remains unproven.

I'd suggest reframing this PR as 'fix: output errors are now surfaced with non-zero exit' rather than claiming it closes #249 outright. The original WSL issue should stay open until someone can reproduce it on actual WSL with specific details:
- Was the project under /mnt/c or /home?
- What does 'uvx pyscn --version' show?
- What was the actual exit code when the report was missing?
- Does .pyscn/reports exist after running?

This way we get a valid fix merged while keeping the WSL investigation alive.

cmd/pyscn/utils_test.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		t.Fatalf("resolveOutputDirectory returned error: %v", err)
+	}
+
+	// Verify it defaults to .pyscn/reports under CWD
+	expectedDir := filepath.Join(tempDir, ".pyscn", "reports")
+	if outputDir != expectedDir {
+		t.Errorf("expected directory %q, got %q", expectedDir, outputDir)
+	}
+}
```
DaisukeYoda, line 88: Since os.Getwd() essentially never fails in practice, this test never exercises the fallback branch it's named after. Let's drop it as it's effectively dead code.

cmd/pyscn/utils_test.go
```diff
@@ -0,0 +1,150 @@
...
+	// This test is disabled pending a config loader fix for relative paths.
+	t.Skip("Skipping - config loader returns relative paths for custom output.directory")
+
+	// This test would verify that when a user sets:
+	//   [output]
+	//   directory = "custom_reports"
+	// in their .pyscn.toml, the output path uses that directory.
+	// The config loader currently doesn't convert relative paths to absolute,
+	// which is a separate bug from #249.
+}
```
DaisukeYoda: Relative paths resolving against CWD looks like intentional POSIX behavior to me, and I don't see an existing issue for this. Could you either open a separate issue describing the expected behavior, or drop this skipped test?

cmd/pyscn/analyze.go
```diff
@@ -170,10 +170,12 @@ func (c *AnalyzeCommand) runAnalyze(cmd *cobra.Command, args []string) error {
 	response, analysisErr := useCase.Execute(ctx, config, args)
 
 	// Generate output even if there were partial failures
+	var outputErr error
 	if response != nil {
 		// Generate output
 		if err := c.generateOutput(cmd, response, args); err != nil {
 			fmt.Fprintf(cmd.ErrOrStderr(), "Warning: Failed to generate output: %v\n", err)
```
DaisukeYoda: This PR changes what propagates to the caller, so Cobra now prints the error itself. The manual fmt.Fprintf above becomes duplicate output. please remove it.

DaisukeYoda (changes requested):
@CatfishGG 

As you and Hassan mentioned, this PR doesn't solve the WSL issue itself but is still a valuable fix. Please take a look at my review comments, and I'd be happy to merge this PR afterward.

Thank you!


(Also, please change Fixes #249 to Refs #249 in the PR description so #249 stays open for further WSL investigation)

CatfishGG (commented):
Thanks for the reviews — addressed all three points in the latest push:

1. **Dropped TestGenerateOutputFilePath_AllowsCustomDirectoryViaConfig** — you're right, the relative path resolution in the config loader is a separate concern and no issue is filed for it. Removed.

2. **Dropped TestResolveOutputDirectory_FallbackToRelative** — agreed, os.Getwd() essentially never fails, making this effectively dead code. Removed.

3. **CI fixes** — removed the duplicate fmt.Fprintf warning now that the error propagates up to cobra, and added `//nolint:errcheck` on the deferred os.Chdir calls. gofmt and go build pass locally. Pushed as debcaf9.

DaisukeYoda:
@CatfishGG 

Could you take a look at my latest review comment? //nolint:errcheck isn't needed here since errcheck isn't enabled in this project's .golangci.yml.

Also, please change Fixes #249 to Refs #249 in the PR description. The Fixes keyword will make GitHub auto-close #249 when this PR is merged, which isn't what we want as discussed above.

CatfishGG:
Thanks for the review! All four points addressed in the latest push (debcaf9):
- Removed TestResolveOutputDirectory_FallbackToRelative
- Removed TestGenerateOutputFilePath_AllowsCustomDirectoryViaConfig
- Removed duplicate fmt.Fprintf warning
- Added //nolint:errcheck on deferred os.Chdir calls

PR is ready for another look whenever you have time.

CatfishGG:
Noted on changing Fixes #249 to Refs #249 in the description, will do. In the meantime, all three review comments are addressed. Ready for another pass whenever you have time!

DaisukeYoda:
@CatfishGG 


Thanks! Blocker on the new tests: they fail locally on macOS because `t.TempDir()` returns `/var/folders/...` but `os.Getwd()` after `os.Chdir` returns the symlink-resolved `/private/var/...`. CI is Ubuntu-only so it passes there.

Could you wrap `tempDir` with `filepath.EvalSymlinks` before `os.Chdir` / building the expected path? Applies to all three tests.

CatfishGG:
Fixed — added `filepath.EvalSymlinks` on `t.TempDir()` in all three tests before `os.Chdir`. Tests pass on Linux, should pass on macOS now too. Pushed as ffb66bb.

DaisukeYoda (approved):
LGTM! Thank you!

PR #441, opened 2026-05-17
fix: change default output directory from .pyscn/reports to reports

### LINKED ISSUE #249: [BUG] `uvx pyscn analyze .` does not generate HTML report on WSL
## Description

When running the following command:

```bash
uvx pyscn analyze .
````

the command completes without any obvious errors, but **no HTML report is generated**.

According to the README, `pyscn analyze` should produce a comprehensive analysis with an HTML report. I expected the output to appear in a directory such as `reports/`, but nothing is created.

## Environment

* OS: Windows 11 (WSL)
* `uv` version: 0.7.3
* Python version: 3.11.12
* `pyscn` version: not installed globally, only using `uvx pyscn` (run on 2025-12-07)
* Installation method: `uvx pyscn` (no global installation)

## Steps to Reproduce

1. Open terminal inside a Python project directory.

2. Run:

   ```bash
   uvx pyscn analyze .
   ```

3. The command finishes without errors.

4. Check the workspace:

   ```bash
   ls
   ls reports
   ```

5. No HTML report or output directory is present.

## Expected Behavior

* An HTML report should be generated after running:

  ```bash
  uvx pyscn analyze .
  ```

* If no `.pyscn.toml` file is present, I expect a **default output directory** to be created automatically (e.g. `./reports/`), containing the report.

## Actual Behavior

* No HTML report is produced.
* No output directory is created.
* No warnings or errors are shown.
* No browser is opened.

## Additional Context

The same behavior occurs in different project directories.
I did not install `pyscn` globally; I only executed via `uvx`.

If there is a required argument or configuration step for report output when using `uvx`, clarifying this in the documentation would be very helpful.

```

aftab1337:
## Summary

The HTML report was being generated to a hidden directory (`.pyscn/reports/`), which caused first-time users to think no report was being created at all. The report path is printed in the output, and `reports/` is now a visible directory that users can actually spot.

## Changes

- **cmd/pyscn/utils.go**: Changed default output dir from `.pyscn/reports` to `reports`
- **internal/config/config.go**: Updated comment to reflect new default
- **internal/config/pyscn_config.go**: Updated comment to reflect new default

## Testing

Built and tested on WSL2 Ubuntu. Verified that the HTML report is now generated in a visible `reports/` directory instead of the hidden `.pyscn/reports/`.

## Fixes

Fixes #249

### DIFF
--- a/cmd/pyscn/utils.go
+++ b/cmd/pyscn/utils.go
@@ -31,14 +31,13 @@ func resolveOutputDirectory(targetPath string) (string, error) {
 	}
 
 	// Default output directory when not specified in config
-	// Use a tool-specific hidden directory under the current working directory
-	// (avoids writing into analyzed source directories by default)
+	// Uses a visible reports/ directory under the current working directory.
 	cwd, err := os.Getwd()
 	if err != nil {
 		// Fallback to relative path if CWD not available
-		return filepath.Join(".pyscn", "reports"), nil
+		return filepath.Join("reports"), nil
 	}
-	return filepath.Join(cwd, ".pyscn", "reports"), nil
+	return filepath.Join(cwd, "reports"), nil
 }
 
 // generateOutputFilePath combines filename generation and directory resolution
@@ -53,7 +52,7 @@ func generateOutputFilePath(command, extension, targetPath string) (string, erro
 
 	// Ensure the directory exists before returning the path. At this point,
 	// outputDir is always non-empty because resolveOutputDirectory provides
-	// a default (e.g., .pyscn/reports under CWD) when config is unset.
+	// a default (e.g., reports/ under CWD) when config is unset.
 	if mkErr := os.MkdirAll(outputDir, 0o755); mkErr != nil {
 		return "", fmt.Errorf("failed to create output directory %s: %w", outputDir, mkErr)
 	}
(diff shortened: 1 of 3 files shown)


aftab1337:
@DaisukeYoda @M-Hassan-Raza thoughts?

DaisukeYoda:
@aftab1337 

Thank you for your suggestion, but tbh I’m hesitant to change the default output directory from `.pyscn/reports` to `reports/`.

In the Python ecosystem, tool-managed output under hidden, tool-owned paths is effectively the norm: `ruff` uses `.ruff_cache/`, `mypy` uses `.mypy_cache/`, `pytest` uses `.pytest_cache/`. Using `.pyscn/reports` follows that convention and avoids creating a visible, generic `reports/` directory in user repositories by default. A visible `reports/` directory is better when users explicitly configure it, but I don’t think the CLI should start writing there implicitly.

I’d prefer to keep `.pyscn/reports` as the default and fix the UX/documentation instead: make the generated report path more prominent and ensure docs clearly say where the default report is written.

DaisukeYoda:
@aftab1337 I am closing this PR for the reason I mentioned above.  Again, I appreciate your verification and suggestion.

PR #452, opened 2026-05-19
feat(analyzer): TF-IDF + Cosine Similarity for Type-2 clone detection (fixes #315)

### LINKED ISSUE #315: Introduce TF-IDF + Cosine Similarity for Type-2 Clone Detection to Reduce False Positives

CatfishGG:
## Summary

TF-IDF weighted cosine similarity for Type-2 clone detection.

### Problem

The previous APTED-based similarity for Type-2 clones used pure tree edit distance. Issue #315 requested TF-IDF weighting to better capture code semantics.

### Solution

1. **New interface methods** on SimilarityAnalyzer: ComputeDistance, ComputeDistanceAndSimilarity
2. **Renamed tree methods** in APTEDAnalyzer: ComputeDistanceTrees / ComputeSimilarityTrees
3. **Added ComputeDistanceAndSimilarityTrees** to APTEDAnalyzer for single-traversal distance+similarity
4. **Eliminated double APTED runs** in compareFragmentsWithClassifier and compareWithAPTED

### Files changed (14 files, net -328 lines)

- Renamed APTED tree methods, added combined ComputeDistanceAndSimilarity methods
- Updated CloneDetector to use the combined method, eliminating 2x APTED cost per clone pair
- Interface updated with new methods: ComputeDistance, ComputeDistanceAndSimilarity
- All analyzers (APTED, Semantic, Structural, Syntactic, Textual) updated

### Test results

All analyzer tests pass.

### Notes

- TF-IDF feature code stripped to empty stub per Yoda review
- Full TF-IDF implementation deferred to PR B
- ComputeSimilarity on SimilarityAnalyzer interface restored to 2-argument signature


DaisukeYoda:
@CatfishGG Hello! 

You didn't have to create new PR and you can use existing one to solve the conflicts. Although it's okay, please close the old PR.

CatfishGG:
Thanks for the heads up Yoda, my bad. Won't happen again - will always use the existing PR to resolve conflicts going forward. And noted on closing the old one, just did.

DaisukeYoda:
Hi @CatfishGG — thanks for picking this back up. I took a closer look and I think the cleanest path forward is to **split this PR in two**. Here's why and what I'm proposing.

### What I found

1. **TF-IDF path is dead code.** `cd.tfidfCalculator` is declared but never assigned (no `NewTFIDFCalculator()` call from `NewCloneDetector`, no `UseTFIDF` plumbing from CLI/config), so `SyntacticSimilarityAnalyzer.ComputeSimilarity` always hits the `calc == nil` branch and falls back to Jaccard. Type-2 behavior is effectively unchanged from main.

   FWIW you actually had this wired correctly in #439 (`newTFIDFCalculatorFromConfig` + `BuildFromCorpus` in `prepareFragments`). It looks like the wiring got lost during the rebase/reissue — worth diffing #439's final state against this branch.

2. **`ToWeightedVector` doesn't apply IDF.** The comment at `tfidf.go:67-69` admits IDF is hardcoded to 1.0. Even if the calculator were wired, the output is just TF-Cosine, not TF-IDF Cosine. The corpus-side `BuildFromCorpus` from #439 needs to come back and feed `ToWeightedVector` with the IDF map.

3. **APTED now runs twice per pair.** `ComputeDistanceAndSimilarity` was removed, replaced by separate `ComputeDistance` + `ComputeSimilarity` calls in `compareFragmentsWithClassifier` / `compareWithAPTED`. Since `ComputeSimilarityTrees` internally calls `ComputeDistanceTrees`, every fragment pair now pays 2× APTED cost. This is a significant regression on large codebases.

4. **`Features` fast path lost.** Main's `SyntacticSimilarityAnalyzer.ComputeSimilarity` short-circuited to `jaccardSimilarity(f1.Features, f2.Features)` when pre-computed features were available (populated in `prepareFragments`). The PR version always re-extracts via `s.extractor.ExtractFeatures(tree)` — extra work on every Type-2 comparison.

### Proposed split

**PR A — pure refactor, no behavior change**
- Rename `APTEDAnalyzer.ComputeDistance`/`ComputeSimilarity` → `ComputeDistanceTrees`/`ComputeSimilarityTrees`
- Add `CodeFragment`-based `ComputeSimilarity(f1, f2, calc)` / `ComputeDistance(f1, f2)` to `SimilarityAnalyzer` interface and update all implementors
- **Keep `ComputeDistanceAndSimilarity`** and keep using it in `compareFragmentsWithClassifier` / `compareWithAPTED`
- **Keep the `Features` fast path** in `SyntacticSimilarityAnalyzer.ComputeSimilarity`
- `TFIDFCalculator` param can be plumbed through as unused (`_ *TFIDFCalculator`) for now, or we can defer the param addition to PR B — your call

This one should be reviewable quickly since it's mechanical.

**PR B — actual TF-IDF, on top of A**
- Restore `BuildFromCorpus` and `newTFIDFCalculatorFromConfig` from #439
- Rework `ToWeightedVector` to take an IDF map (or call `ComputeIDF`) so the output is actually TF × IDF, not just TF
- Wire `UseTFIDF` from CLI flag / config through `CloneDetectorConfig`
- Add a fixture test that demonstrates the FP reduction claim — e.g. two `@dataclass`-decorated classes with different business logic should drop below `Type2Threshold` with TF-IDF on but exceed it with Jaccard

The fixture test is important — it's the only thing that can verify the PR delivers on the issue's promise.

Let me know if you'd like to take this on, or if you'd prefer to hand it off — either is fine. If you go for it, PR A is the lower-risk start.

CatfishGG:
Closing in favor of a split approach. The TF-IDF work (commits 2572059 through 82345d0) will be resubmitted as a separate PR after addressing review feedback. PR A (this one) contains only the bug fixes for findings #3 and #4.

DaisukeYoda:
Thanks for the perf fixes. But this isn't pure refactor yet: `tfidf.go`, `tfidf_test.go`, `UseTFIDF` flag, and `tfidfCalculator` field are TF-IDF feature code. Please move them to PR B.

The `*TFIDFCalculator` interface param can stay if you keep an empty stub here:

```go
// tfidf.go
package analyzer
type TFIDFCalculator struct{}
```

Also drop `Closes #315` from the PR body so merging doesn't auto-close the issue.

CatfishGG:
My bad. I mistakenly changed the name. :skull:

PR #454, opened 2026-05-19
Fix large-tree APTED distance handling

M-Hassan-Raza:
This fixes the optimized APTED path for large trees so label differences and configured cost models are still respected.

It also adds coverage for large trees, ignored identifiers, weighted rename costs, shifted and reversed siblings, and budget fallback behavior.

Validation:
go test ./...
go test ./internal/analyzer -run 'TestAPTEDAnalyzer_(LargeTreesPreserveLabelDistance|SameShapeDistanceMatchesExactAPTED)' -bench BenchmarkAPTED_LargeTreeLabelDistance -benchtime=100ms -count=1

### DIFF
--- a/internal/analyzer/apted.go
+++ b/internal/analyzer/apted.go
@@ -3,6 +3,7 @@ package analyzer
 import (
 	"math"
 	"sort"
+	"strings"
 )
 
 // APTEDAnalyzer implements the APTED (All Path Tree Edit Distance) algorithm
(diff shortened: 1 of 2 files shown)


M-Hassan-Raza:
@DaisukeYoda I experimented with a few variations but finally landed on this as the perfect balance (imo) for speed and accuracy. Please offer your suggestions if you have any different ideas. This PR is a little subjective than previous ones 😁

DaisukeYoda (approved):
@M-Hassan-Raza 

Thanks a lot. I reviewed it and this looks good.

Given the current false-positive and performance pressure around large trees, this direction makes sense to me. I agree this is a subjective tradeoff, but the label/profile distance improvement looks like a practical step.

One thing I’d like us to make explicit is that the large-tree path is now more of a clone-detection heuristic than exact APTED distance. The same-shape fast path can overestimate exact APTED in some cases, so documenting that assumption in the code or tests would help future maintainers.

Thanks!

M-Hassan-Raza:
@DaisukeYoda Thanks, agreed. I’ll add a short note in the code/tests to make that assumption explicit: the large-tree path is optimized for clone-detection signal, not exact APTED in every reorder case, and the same-shape fast path can overestimate exact distance. I’ll push that cleanup in a separate PR shortly

PR #466, opened 2026-05-22
Fix complexity metric source consistency

M-Hassan-Raza:
This tightens the complexity metrics so public counters come from the right source.

When a CFG has its original AST node, the analyzer now uses parser-backed counts for if/elif statements, loops, exception handlers, and match cases. CFG-only callers still keep the existing fallback behavior.

This also fixes match complexity so a match block does not get counted once as a synthetic CFG decision and again through its cases.

Added focused regression coverage and updated the complexity docs to match the analyzer behavior.

Validation:
go test ./...


DaisukeYoda (approved):
Thanks, LGTM! Nice catch on this bug!

PR #467, opened 2026-05-22
Clarify large-tree APTED heuristic

M-Hassan-Raza:
Clarifies that the large-tree APTED path is a bounded clone-detection heuristic, not exact APTED for every reorder case.


DaisukeYoda (approved):
Thanks for the follow-up! The clarification matches the PR #454 behavior. LGTM!

PR #474, opened 2026-05-23
Fix CBO dependency identity

M-Hassan-Raza:
This fixes CBO false positives caused by counting repeated dependency sightings, type-system bases, self references, and leaked import aliases as real coupling.

CBO now tracks unique dependency identities by dependency kind. Type-system imports from typing, typing_extensions, and abc are preserved in base class metadata but no longer inflate CBO. IncludeImports still controls normal stdlib imports, and qualified project dependencies such as models.Widget or models.Protocol continue to count.

Added analyzer and service-level regression coverage for the public output contract.

Validation:
go test ./internal/analyzer ./service -count=1
go test ./... -count=1

### DIFF
--- a/internal/analyzer/cbo.go
+++ b/internal/analyzer/cbo.go
@@ -3,6 +3,7 @@ package analyzer
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/ludo-technologies/pyscn/domain"
@@ -73,6 +74,36 @@ type CBOAnalyzer struct {
 	regexCache       map[string]*regexp.Regexp // pattern -> compiled regex cache
 }
 
+type cboDependencyKind int
+
+const (
+	dependencyKindInheritance cboDependencyKind = iota
+	dependencyKindTypeHint
+	dependencyKindInstantiation
+	dependencyKindAttributeAccess
+	dependencyKindImport
+)
+
+type cboDependencies struct {
+	all               map[string]bool
+	inheritance       map[string]bool
+	typeHints         map[string]bool
+	instantiations    map[string]bool
+	attributeAccesses map[string]bool
+	imports           map[string]bool
+}
+
+func newCBODependencies() *cboDependencies {
+	return &cboDependencies{
+		all:               make(map[string]bool),
+		inheritance:       make(map[string]bool),
+		typeHints:         make(map[string]bool),
+		instantiations:    make(map[string]bool),
+		attributeAccesses: make(map[string]bool),
+		imports:           make(map[string]bool),
+	}
+}
+
 // NewCBOAnalyzer creates a new CBO analyzer
 func NewCBOAnalyzer(options *CBOOptions) *CBOAnalyzer {
 	if options == nil {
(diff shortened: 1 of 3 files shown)


internal/analyzer/cbo.go
```diff
@@ -326,19 +344,15 @@ func (a *CBOAnalyzer) analyzeInstantiationAndAccess(classNode *parser.Node, depe
...
 							} else if _, isClass := allClasses[className]; isClass {
 								// Known local class (instantiation)
-								dependencies[className] = true
-								result.InstantiationDependencies++
+								a.addDependency(dependencies, className, dependencyKindInstantiation)
 							} else if a.options.IncludeBuiltins && a.builtinTypes[className] {
 								// Builtin type (only if explicitly enabled)
-								dependencies[className] = true
-								result.InstantiationDependencies++
+								a.addDependency(dependencies, className, dependencyKindInstantiation)
```
DaisukeYoda, line 349: The caller branches on `isImportedDependency` and then `addDependency` checks it again, overriding the kind. The caller's branch is dead logic, so please drop it and let `addDependency` own the routing.

internal/analyzer/cbo.go
```diff
@@ -497,24 +537,133 @@ func (a *CBOAnalyzer) shouldIncludeDependency(className string) bool {
...
+
+func (a *CBOAnalyzer) sameDependencyIdentity(className, ownerClass string) bool {
+	if className == ownerClass {
+		return true
+	}
+	if imported, exists := a.importedNames[className]; exists {
+		return imported == ownerClass
+	}
+	return false
+}
```
DaisukeYoda, line 587: comparison rarely matches `importedNames[className]` values are usually qualified
("pkg.Widget") while `ownerClass` is the bare name ("Widget"), so this branch almost never fires. Compare against the last segment of the imported value instead, and add a test for `from pkg import
Widget` inside `class Widget:`.

internal/analyzer/cbo.go
```diff
@@ -497,24 +537,133 @@ func (a *CBOAnalyzer) shouldIncludeDependency(className string) bool {
...
+
+	dependencyName := className
+	if imported, exists := a.importedNames[className]; exists {
+		dependencyName = imported
+	} else if !strings.Contains(className, ".") {
+		return false
+	}
+
+	return a.typeSystemRoot(dependencyName) && a.isTypeSystemName(dependencyName)
+}
```
DaisukeYoda, line 610: `isTypeSystemName`'s whitelist misses common typing names (`MutableMapping`, `Awaitable`, `Coroutine`, `Generator`, `Annotated`, …), so inheriting them still inflates CBO.

How about dropping the list and filtering by import root: if the name resolves to `typing` / `typing_extensions` / `abc`, treat as type-system?

DaisukeYoda (changes requested):
Thanks as always! Please take a look at some of my comment.

M-Hassan-Raza:
@DaisukeYoda Appreciate the solid in-depth review! Should be solid now hopefully. If any more blockers remain please lmk :)

DaisukeYoda (approved):
LGTM!

PR #495, opened 2026-05-30
Fix analyzer import and report contracts

M-Hassan-Raza:
Fixes analyzer contract issues across module dependency analysis, parser traversal, config loading, and HTML report output.

This includes correct comprehension fields, import and re-export resolution, stub module handling, source vs module include defaults, escaped report strings, explicit include pattern handling, and safer trusted metric HTML.

Validated with go test -count=1 ./..., go vet ./..., and go test -race -count=1 ./internal/analyzer ./service.


DaisukeYoda (approved):
LGTM 👍  All solid improvements. Thanks for the careful fixes and thorough tests!

PR #498, opened 2026-05-30
Fix DFA compound statement use contracts

M-Hassan-Raza:
Fixes DFA extraction for compound statements so header uses, scoped definitions, defaults, and match patterns are attributed to the correct CFG blocks.

Keeps typed parameter annotations on Arg.Right and default expressions on Arg.Value, with CBO and dependency checks using the same contract.

Validated with go test ./..., go vet ./..., go test -race ./internal/analyzer -count=1, and golangci-lint v1.64.8.


DaisukeYoda (approved):
LGTM!

PR #502, opened 2026-05-31
Fix module import branch traversal

M-Hassan-Raza:
Module dependency analysis missed imports stored outside the AST Body field, including imports in else, except, and finally branches. This makes module analysis use the parser traversal contract, collects module facts in one pass, and keeps TYPE_CHECKING filtering branch-aware.

Validation: go test ./internal/analyzer and go test ./...


DaisukeYoda (approved):
LGTM! Thank you!

Issue #488, opened 2026-05-26
[BUG][auto] clone-fragment-id-type-zero: fragment id/type/hash always zero in clone group output


DaisukeYoda:
## Bug description

In clone detection JSON output, every clone **fragment** within a clone group has `"id": 0` and `"type": 0`, regardless of the actual clone type or fragment identity. The **group-level** `type` field is correctly populated (1, 2, or 4), but individual fragment fields are never set.

This means:
- Fragments within a group cannot be uniquely identified (all share `id: 0`)
- Fragment-level clone type is always `0` (unset) even when the group type is Type-1, Type-2, or Type-4
- The `hash` field on fragments is always an empty string `""`

## Repro

Analyze any Python repository with clone detection:

```bash
pyscn analyze --json <repo-path>
```

Inspect `clone.clone_groups[*].clones[*]` in the JSON output. Every fragment will have:

```json
{
  "id": 0,
  "type": 0,
  "hash": "",
  ...
}
```

While the parent group correctly shows:

```json
{
  "id": 16,
  "type": 1,
  "similarity": 1.0,
  ...
}
```

## Expected behavior

Each fragment should have:
- A unique `id` within its group (or globally)
- A `type` matching the group's clone type (or the fragment's own classification)
- A populated `hash` if the detection algorithm computed one

## Actual output

Tested on `laurentS/slowapi` (commit `5e2b9c1`), 13 clone groups / 28 pairs — **every** fragment across all groups has `id: 0`, `type: 0`, `hash: ""`.

```json
{
  "id": 16,
  "clones": [
    {"id": 0, "type": 0, "hash": "", "location": {"file_path": "slowapi/extension.py", "start_line": 404, "end_line": 408}, "size": 13, "line_count": 5},
    {"id": 0, "type": 0, "hash": "", "location": {"file_path": "slowapi/extension.py", "start_line": 456, "end_line": 460}, "size": 13, "line_count": 5}
  ],
  "type": 1,
  "similarity": 1
}
```

## Priority rationale

P2 — visible but bounded inaccuracy: the group-level type is correct so the gist of the detection is right, but fragment-level fields are internally inconsistent (always zero/empty), breaking downstream consumers that need per-fragment identity or type info.

## Version

- pyscn v1.22.3-2-g94ce860
- Found via the FP-audit skill in repo `laurentS/slowapi@5e2b9c1`

PR #503, opened 2026-05-31
fix(clones): populate fragment id and type in clone group output

StressTestor:
## Summary

Clone-group fragments in the analysis output (`clone.clone_groups[*].clones[*]`) were emitted with `id: 0` and `type: 0` for **every** fragment, even though the parent group carried the correct `id` and clone `type`. That left fragments within a group indistinguishable from one another and untyped.

Root cause: `convertCloneGroupsToDomain` (`service/clone_service.go`) built each `domain.Clone` with only `Location`/`Size`/`LineCount` and never set `ID` or `Type`. The sibling converter `convertFragmentsToDomainClones` already sets `ID: i + 1`; this aligns the group path with it and inherits the group's clone type per fragment.

## Type of Change

## Related Issues
Refs #488

> **Scope note:** #488 also reports `hash: ""` on fragments. That's a *separate* root cause — `analyzer.CodeFragment.Hash` is never populated anywhere in the detector (clone detection uses MinHash/LSH signatures, not a stored per-fragment hash), so the field is dead at the source. Populating it well needs a design decision (hash algorithm + whether Type-1 whitespace/comment normalization applies), which felt like a maintainer call rather than something to invent here. This PR forwards `fragment.Hash` (so output tracks the field once it's populated) but doesn't compute it — happy to follow up on hash separately. Used `Refs` rather than `Closes` so the hash part stays tracked on #488.

## Changes
- `service/clone_service.go`: in `convertCloneGroupsToDomain`, set per-fragment `ID` (`i + 1`, unique within the group, matching `convertFragmentsToDomainClones`) and `Type` (inherited from the group's clone type); forward `Hash`.
- `service/clone_service_content_test.go`: add `TestCloneService_convertCloneGroupsToDomain_FragmentIDAndType` asserting fragments get sequential ids and the group's clone type.

## Testing
- [x] `make test` (`go test ./...`) passes locally
- [x] `go test -race ./service/...` passes locally
- [x] `make lint` passes locally (`go vet ./...` + `golangci-lint run`, both clean)
- [x] Added a regression test

Manual before/after on a two-function Type-2 clone fixture:

```
# before
clones: [ {id:0, type:0}, {id:0, type:0} ]
# after  (group: id:0, type:2)
clones: [ {id:1, type:2}, {id:2, type:2} ]
```

## Documentation
- [x] No doc changes needed — the output schema fields already exist; this populates them.

### DIFF
--- a/service/clone_service.go
+++ b/service/clone_service.go
@@ -382,8 +382,10 @@ func (s *CloneService) convertCloneGroupsToDomain(cloneGroups []*analyzer.CloneG
 		}
 
 		// Convert fragments to clones
-		for _, fragment := range group.Fragments {
+		for i, fragment := range group.Fragments {
 			clone := &domain.Clone{
+				ID:   i + 1,
+				Type: s.convertCloneType(group.CloneType),
 				Location: &domain.CloneLocation{
 					FilePath:  fragment.Location.FilePath,
 					StartLine: fragment.Location.StartLine,
@@ -391,6 +393,7 @@ func (s *CloneService) convertCloneGroupsToDomain(cloneGroups []*analyzer.CloneG
 					StartCol:  fragment.Location.StartCol,
 					EndCol:    fragment.Location.EndCol,
 				},
+				Hash:      fragment.Hash,
 				Size:      fragment.Size,
 				LineCount: fragment.LineCount,
 			}
--- a/service/clone_service.go
+++ b/service/clone_service.go
@@ -162,9 +162,9 @@ func (s *CloneService) buildCloneResponse(
 	clonePairs, cloneGroups := detector.DetectClonesWithLSH(ctx, allFragments)
 
 	// Convert to domain objects
-	domainClones := s.convertFragmentsToDomainClones(allFragments)
+	domainClones, fragmentIDs := s.convertFragmentsToDomainClones(allFragments)
 	domainClonePairs := s.convertClonePairsToDomain(clonePairs, req.ShowContent)
-	domainCloneGroups := s.convertCloneGroupsToDomain(cloneGroups, req.ShowContent)
+	domainCloneGroups := s.convertCloneGroupsToDomain(cloneGroups, req.ShowContent, fragmentIDs)
 
 	// Filter results based on request criteria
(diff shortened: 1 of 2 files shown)


service/clone_service.go
```diff
@@ -382,15 +382,18 @@ func (s *CloneService) convertCloneGroupsToDomain(cloneGroups []*analyzer.CloneG
 		}
 
 		// Convert fragments to clones
-		for _, fragment := range group.Fragments {
+		for i, fragment := range group.Fragments {
```
DaisukeYoda: `i + 1` resets for each group, so fragment IDs are not unique within the response. It also breaks the link to the same fragment in top-level `clones[]` (for example, `id=3` there may become `id=1` here).

DaisukeYoda (changes requested):
Thank you for submitting this PR. Please take a look at my review comment.

service/clone_service.go
StressTestor: Good catch, thanks. Fixed in bdd6e1d.

`convertFragmentsToDomainClones` now returns a `fragment -> id` map (the same 1-based id it assigns to the top-level `clones[]`), and `convertCloneGroupsToDomain` reuses it. So a group fragment carries its response-wide id and matches the same fragment in `clones[]`, instead of resetting `1..n` per group.

Regression test now assigns non-sequential response-wide ids (3 and 7) and asserts the group surfaces those exact ids; also verified end-to-end that group fragment ids link back to `clones[]`.

DaisukeYoda:
@StressTestor could you check the current status?

StressTestor:
hey @DaisukeYoda — thanks for the nudge. the review fix is in: `convertFragmentsToDomainClones` now returns a `fragment -> id` map (the 1-based response-wide index, the same id used in top-level `clones[]`), and `convertCloneGroupsToDomain` reuses it — so a group fragment carries its response-wide id instead of resetting `1..n` per group, and it links back to the same fragment in `clones[]`.

i also just rebased onto current main (it had drifted into a merge conflict) — clean now, `go build ./...` and the `service` tests pass. should be ready for another look.

DaisukeYoda (approved):
LGTM!

PR #537, opened 2026-06-10
fix(parser): preserve every elif in a chain so complexity isn't undercounted (#511)

### LINKED ISSUE #511: [ENH][auto] elif-chain-truncation-complexity-undercount: parser keeps only the last elif of a chain, undercounting complexity
**Track**: false-negative (complexity under-counting caused by a parser bug)

## Summary

pyscn's Python parser keeps **only the last `elif` clause** of a multi-`elif` chain and silently discards all the preceding ones before the CFG is built. As a result, cyclomatic complexity is understated by roughly *N − 1* for any function with an *N*-branch `if/elif/…/elif` dispatch. Functions that are genuinely hard to test (protobuf deserializers, AMQP frame encoders/decoders, type-switch dispatchers) are either reported far below their true complexity or — when the deflated value drops under the reporting threshold — omitted from the report entirely.

## Rationale (user value)

Long `if/elif` dispatch chains are exactly the kind of branch-dense code pyscn exists to surface. Today a function with a 10-branch `elif` chain registers as complexity ≈ 3 instead of ≈ 11, giving users false confidence that switch-like dispatchers are simple. This is a correctness bug in the core complexity metric, so it affects every downstream score (risk level, health score, `pyscn check` gating).

## Root cause (pyscn symbols, verbatim)

`buildIfStatement` in `internal/parser/ast_builder.go` (function at `internal/parser/ast_builder.go:301`) iterates over the `if_statement` node's children that carry the tree-sitter field name `alternative`. Python's tree-sitter grammar emits **every** `elif_clause` as a direct sibling of the parent `if_statement`, each tagged `field == "alternative"` (verified empirically: `if/elif/elif/else` yields `child=elif_clause, elif_clause, else_clause`, all with field `alternative`). The loop assigns the elif to a single variable and **overwrites it on each iteration** (`elifNode = alt` at `internal/parser/ast_builder.go:329`), so only the final `elif_clause` survives; the earlier ones never enter the AST and produce no CFG decision points. `attachElseToElifChain` (`internal/parser/ast_builder.go:352`) correctly attaches a trailing `else`, but there is no equivalent chaining of one `elif` to the next.

The existing tests `TestProcessIfStatementElif` / `TestComplexElifScenarios` pass because they assert only lower bounds on block counts, which remain satisfied when intermediate `elif`s are dropped.

## Proposed change

In the `alternative`-scanning loop, **chain** successive `elif` clauses instead of overwriting. Each new `elif_clause` should be linked as the `Orelse` of the previous one (e.g. `prevElif.Orelse = []*Node{newElifNode}`), reconstructing the nested `if/else` ladder that Python semantics imply, so the CFG builder creates a decision point for every branch. Note: do **not** reuse `attachElseToElifChain(elifNode, newAlt)` for elif-to-elif linking — its second argument is an `else_clause`, not an `elif`; chain via the `Orelse` slot instead. After the fix, add a complexity assertion (not just a block-count lower bound) on a 3+ `elif` function so the regression is caught.

## Evidence (aleksul/repid @ 12a9f64)

Run: `pyscn analyze --json` (v1.22.7). From the report:

- **`repid/connections/amqp/_uamqp/_decode.py:417` `_construct_message`** — a `while` loop containing four guard `if`s plus an 8-branch `elif` descriptor dispatch (true McCabe ≈ 16). The function carries the author's own `# noqa: C901, PLR0912` marker, confirming it is known-complex. pyscn reports:
  ```json
  { "Name": "...bytes_to_performative / _construct_message", "Metrics": { "Complexity": 7, "IfStatements": 5 } }
  ```
  Only the four plain `if`s plus the single surviving (last) `elif` are counted — `IfStatements: 5`, complexity `7` instead of ~16.

- **`repid/connections/pubsub/protocol/proto.py:506` `StreamingPullRequest.deserialize`** — `while` loop with an 8-branch `elif` field-number dispatch (true McCabe ≈ 10). **Absent entirely** from the 115-function report: its deflated complexity falls below the `min_complexity` reporting threshold.

- **`repid/connections/amqp/_uamqp/_encode.py:731` `_encode_unknown`** — `if` + 9 `elif` `isinstance` type dispatch (true McCabe ≈ 11). **Absent** from the report for the same reason.

## Minimal synthetic repro

```python
def classify(x):
    if x == 1:
        return "a"
    elif x == 2:
        return "b"
    elif x == 3:
        return "c"
    elif x == 4:
        return "d"
    elif x == 5:
        return "e"
    else:
        return "z"
```
True cyclomatic complexity = 6 (5 decision points + 1). With the bug, only the `x == 5` branch is retained, so pyscn computes ≈ 3.

## Impact / effort / confidence

- **Impact**: high — corrupts the core complexity metric and hides genuinely complex dispatchers.
- **Effort**: small — localized to the `alternative` loop in `buildIfStatement`.
- **Confidence**: high.

## Verification

Survived adversarial refutation — the overwrite was confirmed in `ast_builder.go:329`; tree-sitter was confirmed to emit all `elif_clause`s as sibling `alternative` children (so a single retained child is wrong); and the three cited functions' under-counts/omissions were confirmed against the JSON report. The only correction was to the *wording* of the fix (chain via `Orelse`, not via `attachElseToElifChain`), applied above.

pyscn version: v1.22.7 (build `v1.22.7-5-g1200c3e`).

Found via the pyscn-enhancement-audit skill in repo `aleksul/repid@12a9f64bf5536b91c0a75004a87eff3e70eb7410`.

StressTestor:
Fixes #511.

pyscn undercounts cyclomatic complexity for any `if/elif/.../else` chain past the first elif. a function that dispatches on several branches gets scored like it only has one.

repro:

```python
def classify(x):
    if x > 100:
        return "huge"
    elif x > 50:
        return "large"
    elif x > 10:
        return "medium"
    elif x > 0:
        return "small"
    else:
        return "non-positive"
```

four conditions, so cyclomatic complexity should be 5. pyscn reports 3.

## root cause

`buildIfStatement` in `internal/parser/ast_builder.go` walks the `alternative` children of an `if_statement` - tree-sitter exposes every `elif_clause` and the `else_clause` as a separate alternative. it did `elifNode = alt` on each pass, so only the last elif survived. the middle elifs were dropped before the CFG builder ran, and the CFG (which already walks an elif chain correctly via `Orelse`) only ever saw `if -> last elif`.

## the fix

collect the elif clauses in order and link each to the next through `Orelse`, so the parent `if` owns the full nested chain. the else attaches to the last elif. the `attachElseToElifChain` helper expected a chain that was never built, so it's now redundant and removed.

## tests

`internal/analyzer/complexity_elif_chain_test.go` asserts cyclomatic complexity across chain lengths, with and without an else. red before the change (the four-condition case reported 3), green after. the existing `MultipleElif` CFG test only checked `cfg.Size() >= 5`, which is why this slipped through. full `go test ./...`, `go vet ./...`, gofmt, and golangci-lint are all clean.

### DIFF
--- a/internal/analyzer/complexity_elif_chain_test.go
+++ b/internal/analyzer/complexity_elif_chain_test.go
@@ -1,0 +1,54 @@
+package analyzer
+
+import "testing"
+
+// TestCalculateComplexityElifChain pins cyclomatic complexity for elif chains.
+//
+// Each if/elif condition is a decision point, so "if + N elif [+ else]" has
+// cyclomatic complexity N+2 (N+1 conditions plus the base of 1). Regression
+// guard for #511: the AST builder kept only the last elif of a chain, so every
+// elif but the last was dropped and complexity was undercounted on long
+// dispatch chains. The existing MultipleElif CFG test only asserted
+// cfg.Size() >= 5, so it stayed green while branches went missing.
+func TestCalculateComplexityElifChain(t *testing.T) {
+	cases := []struct {
+		name string
+		src  string
+		want int
+	}{
+		{
+			name: "if_else",
+			src:  "\nif x > 1:\n    print(1)\nelse:\n    print(0)\n",
+			want: 2,
+		},
+		{
+			name: "if_one_elif_else",
+			src:  "\nif x > 1:\n    print(1)\nelif x > 2:\n    print(2)\nelse:\n    print(0)\n",
+			want: 3,
+		},
+		{
+			name: "if_three_elif_else",
+			src:  "\nif x > 100:\n    print(1)\nelif x > 50:\n    print(2)\nelif x > 10:\n    print(3)\nelif x > 0:\n    print(4)\nelse:\n    print(0)\n",
+			want: 5,
+		},
+		{
+			name: "if_five_elif_no_else",
+			src:  "\nif x == 1:\n    print(1)\nelif x == 2:\n    print(2)\nelif x == 3:\n    print(3)\nelif x == 4:\n    print(4)\nelif x == 5:\n    print(5)\nelif x == 6:\n    print(6)\n",
+			want: 7,
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			ast := parseSource(t, c.src)
+			cfg, err := NewCFGBuilder().Build(ast)
+			if err != nil {
+				t.Fatalf("Build CFG: %v", err)
+			}
+			got := CalculateComplexity(cfg).Complexity
+			if got != c.want {
+				t.Errorf("cyclomatic complexity = %d, want %d", got, c.want)
+			}
+		})
+	}
+}
--- a/internal/parser/ast_builder.go
+++ b/internal/parser/ast_builder.go
@@ -314,9 +314,12 @@ func (b *ASTBuilder) buildIfStatement(tsNode *sitter.Node) *Node {
 		}
 	}
 
-	// Get alternatives (else/elif) - there may be multiple with the same field name
-	// Tree-sitter can have both elif_clause and else_clause as "alternative"
-	var elifNode *Node
+	// Get alternatives (else/elif). Tree-sitter exposes every elif_clause and
+	// the else_clause as separate "alternative" children of the same
+	// if_statement. Collect the elif clauses in order so the whole chain is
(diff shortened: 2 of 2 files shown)


DaisukeYoda (approved):
Thanks. LGTM!

PR #558, opened 2026-06-15
fix(lcom): count self.<property> reads as method-call edges

### LINKED ISSUE #545: [BUG][auto] lcom4-missed-self-property-call-edge: property reads (self.prop) don't create LCOM4 call edges, splitting connected methods
## Bug

LCOM4 treats `self.xxx()` call expressions as intra-class method-call edges (`extractMethodCalls` in `internal/analyzer/lcom.go`), but a bare attribute read `self.xxx` that resolves to a `@property` of the same class creates no edge. At runtime that read *is* a method invocation via the descriptor protocol, so methods that consume a property get split into a separate connected component and LCOM4 is overcounted.

## Repro

```python
class Config:
    def __init__(self):
        self._x = 1

    @property
    def value(self):
        return self._x

    def use(self):
        return self.value + 1
```

```
pyscn analyze --select lcom repro.py
```

## Expected

`LCOM4 = 1`: `use` invokes the `value` property, and `value` shares `self._x` with `__init__` — all three methods are connected.

## Actual (v1.24.0-7-gd0229fd)

```json
{"Name": "Config", "LCOM4": 2, "MethodGroups": [["__init__", "value"], ["use"]]}
```

`self.value` inside `use` is recorded as an instance-variable access on a variable named `value` that no other method touches, so `use` is isolated.

## Real-world hits (sigstore/sigstore-python@54ca1836)

- `sigstore/_internal/rekor/client.py:217` — `RekorClient` reported LCOM4=4 with `create_entry` isolated, but its body is `return self.log.entries.post(request)` where `log` is a `@property` of the class. The property edge would merge it into the `[__init__, _session, log]` component.
- `sigstore/_internal/oidc/oauth.py:240` — `_OAuthRedirectServer` reported LCOM4=4 with `base_uri` and `redirect_path` each isolated, but `redirect_uri` reads `self.base_uri + self.redirect_path` and `auth_endpoint` reads `self.redirect_uri`. With property edges, LCOM4 should be 2 (only `auth_request_path`, which returns a literal, is genuinely isolated).

## Possible direction

After collecting methods, union any method whose accessed "instance variable" name matches a same-class method/property name (i.e. treat `self.X` as a call edge when `X` is a method of the class). This is the same family as the previously fixed missed-edge bugs #412 (with-statements) and #413 (f-strings).

## Priority rationale

P2: bounded inaccuracy — LCOM4 grouping is provably wrong for property-using classes, but the metric is in the right ballpark and the report is otherwise usable.

Found via the FP-audit skill in repo `sigstore/sigstore-python@54ca18368c665a27b022cf4faaa76abfacbafd6f`.

StressTestor:
## what

a bare `self.<prop>` read of a `@property` now creates an LCOM4 method-call edge to that property, instead of being counted as an instance-variable access.

Fixes #545.

## why

LCOM4 treats `self.xxx()` calls as intra-class edges, but a bare `self.xxx` read that resolves to a `@property` created no edge. so a method that only consumes a property got split into its own connected component and LCOM4 was overcounted.

```python
class Config:
    def __init__(self):
        self._x = 1

    @property
    def value(self):
        return self._x

    def use(self):
        return self.value + 1
```

- before: `LCOM4 = 2`  (`[["__init__", "value"], ["use"]]`)
- after: `LCOM4 = 1`  (`use` connects to `value` via the property call; `value` shares `self._x` with `__init__`)

at runtime that read invokes the getter through the descriptor protocol, so the consumer really is connected.

## fix

in `collectMethods` (`internal/analyzer/lcom.go`): collect the class's `@property` names, then reclassify a bare `self.<prop>` read from an instance-variable access into a call edge, so the existing union pass connects the consumer. the new `collectPropertyNames`/`isProperty` helpers mirror the existing `isClassOrStaticMethod` + `getDecoratorName` handling.

## scope

matches `@property` (the bare decorator name) only, consistent with how the analyzer already resolves decorators. other property-like descriptors (`cached_property`, or dotted forms like `@functools.cached_property`) are out of scope, the same limitation the existing classmethod/staticmethod detection has.

## tests

`internal/analyzer/lcom_test.go`: a class where `use` consumes a property without sharing the underlying field, so LCOM4 must be 1. the existing "class with property included" case stays green - it connects via the shared field, a different path.

### DIFF
--- a/internal/analyzer/lcom.go
+++ b/internal/analyzer/lcom.go
@@ -224,6 +224,11 @@ func (a *LCOMAnalyzer) collectMethods(classNode *parser.Node, declaredFields map
 	calls := make(map[string]map[string]bool)
 	excluded := 0
 
+	// A bare `self.<name>` read of a @property is a method invocation via the
+	// descriptor protocol, not a field access. Collect property names first so
+	// such reads can be recorded as call edges instead of instance variables.
+	propertyNames := a.collectPropertyNames(classNode)
+
 	for _, node := range classNode.Body {
 		if node == nil {
 			continue
@@ -252,6 +257,18 @@ func (a *LCOMAnalyzer) collectMethods(classNode *parser.Node, declaredFields map
 		// Extract self.xxx() method calls
 		methodCalls := make(map[string]bool)
 		a.extractMethodCalls(node, methodCalls)
+
+		// Reclassify bare `self.<prop>` reads: a property access invokes the
+		// getter via the descriptor protocol, so it is a call edge to that
+		// property, not an instance-variable access. Leaving it as a variable
+		// isolates the consuming method and overcounts LCOM4.
+		for name := range propertyNames {
+			if name != node.Name && vars[name] {
+				methodCalls[name] = true
+				delete(vars, name)
+			}
+		}
+
 		calls[node.Name] = methodCalls
 	}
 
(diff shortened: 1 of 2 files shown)


DaisukeYoda (approved):
LGTM! thanks!

PR #562, opened 2026-06-18
fix(cbo): don't count a class defined within a class as coupling

### LINKED ISSUE #547: [BUG][auto] cbo-locally-defined-inner-class-counted: class defined inside own method counted as CBO dependency
## Bug description

CBO counts a class defined *inside the analyzed class's own method* as an external coupling dependency. A locally-scoped helper class is part of the class's own implementation, not coupling to another object, so it inflates `CouplingCount` and pollutes `DependentClasses`.

## Repro

Minimal synthetic case:

```python
class Outer:
    def parse(self):
        def grouper():
            class Helper:
                def __init__(self):
                    self.last = None
            return Helper()
        return grouper()
```

Run `pyscn analyze --json --select cbo` on this file. `Outer`'s `DependentClasses` includes `Helper` and `CouplingCount` is incremented for it.

## Expected behavior

Classes defined within the analyzed class's own body or methods (including inside nested functions) should not count as coupling dependencies. CBO measures coupling *between* objects; a local helper class is internal implementation.

## Actual output

From PGPy (`pgpy/pgp.py`), `PGPKey` is reported with `CouplingCount: 18` and `PktGrouper` in its dependency list:

```json
{"Name": "PGPKey", "FilePath": "pgpy/pgp.py", "StartLine": 1296, "EndLine": 2676,
 "Metrics": {"CouplingCount": 18, "ImportDependencies": 14, "InstantiationDependencies": 4,
 "DependentClasses": ["Armorable", ..., "PktGrouper", ...]}}
```

`PktGrouper` is defined at `pgpy/pgp.py:2622` inside the nested function `pktgrouper()` within `PGPKey.parse()` (lines 2598-2676, inside the `PGPKey` extent 1296-2676 that pyscn itself reports). The real external coupling count is 17.

## Environment

- pyscn `v1.24.0-7-gd0229fd`

## Priority rationale

P2: visible but bounded inaccuracy — the coupling count is off by the number of local helper classes and the gist of the finding is still right.

---
Found via the FP-audit skill in repo `SecurityInnovation/PGPy@30a757181ab02f918a94f8549f354d93639b95e6`.

StressTestor:
## what

CBO counts a class defined inside the analyzed class's own method (or a nested function within it) as an external coupling dependency, which inflates `CouplingCount` and pollutes `DependentClasses`.

```python
class Outer:
    def parse(self):
        def grouper():
            class Helper:
                def __init__(self):
                    self.last = None
            return Helper()
        return grouper()
```

`Outer` gets `CouplingCount: 1` with `DependentClasses: ["Helper"]`, even though `Helper` is its own internal helper. In the issue's real case, `PGPKey` reported `CouplingCount: 18` with `PktGrouper` (defined inside `PGPKey.parse()`) in its list; the real count is 17. This is #547.

## why

`collectClasses` builds a flat name to node map with no scope information, and `analyzeClass` counts any referenced class as coupling. A class's own locally-defined helper is part of its implementation, not coupling between objects.

## fix

After collecting a class's dependencies, drop any whose name is defined within that class's own subtree (its body, methods, and nested functions). A new `collectNestedClassNames` walks the class node (excluding the class itself), and a `remove` helper clears the name from every dependency map so the aggregate and per-kind counts stay consistent.

The exclusion is name-based and scoped per class: it only affects the class that owns the nested definition, so a sibling class's genuine dependency on a same-named top-level class still counts.

## tests

`TestCBOAnalyzer_LocalHelperClassNotCounted`: a class with a method-local helper (inside a nested function) plus a genuine external dependency. The helper is excluded while the external sibling still counts. It fails before the fix (`CouplingCount` 2, `["External", "Helper"]`) and passes after. Full suite green.

Fixes #547.


internal/analyzer/cbo.go
```diff
@@ -197,6 +208,13 @@ func (a *CBOAnalyzer) analyzeClass(classNode *parser.Node, filePath string, allC
 	// 3. Analyze instantiation and attribute access
 	a.analyzeInstantiationAndAccess(classNode, dependencies, result, allClasses)
 
+	// Exclude classes defined within this class's own body or methods
+	// (including nested functions): a locally-scoped helper class is part
+	// of the class's implementation, not coupling between objects. See #547.
+	for name := range a.collectNestedClassNames(classNode) {
+		dependencies.remove(name)
+	}
```
DaisukeYoda: Excluding nested classes is the right direction, but the current implementation removes dependencies by name only. That can also remove a legitimate dependency on a same-named class from a different scope.

For example, if `Outer.parse()` defines a local `Helper` class, but `Outer.build()` calls a top-level `Helper()`, the `build()` call should still count as coupling. With this change, `Helper` is removed from `Outer`'s dependencies just because a nested class with the same name exists somewhere under `Outer`.

The exclusion should be scope-aware: only remove a dependency when the reference actually resolves to the nested class definition, not whenever the dependency name matches a nested class name.

DaisukeYoda (changes requested):
@StressTestor Thanks for this PR! Please take a look at my review comment,

StressTestor:
Good catch, thanks. You're right that removing by name also drops a legitimate cross-scope dependency.

I've reworked it to be scope-aware: a reference is now excluded only when it actually resolves, per Python scoping, to a class defined within the analyzed class. A function-local class is visible only within its own function's subtree, and a class defined in the class body isn't visible from its methods. So your example now behaves correctly: the local `Helper` in `parse()` is excluded, but the top-level `Helper()` call in `build()` still counts.

Added a test for exactly that case (`SameNameClassInDifferentScopeStillCounts`) alongside the original. Pushed.

DaisukeYoda:
@StressTestor 

The scope-aware exclusion now fixes the same-name class case for instantiations and attribute access, but type annotations still bypass that logic.

For example:

```python
class Outer:
    def parse(self):
        class Helper:
            pass
        item: Helper = Helper()
        return item
```

`Helper()` is correctly excluded as a local helper reference, but `item: Helper` is still added through `extractTypeAnnotationDependencies`, so `Outer` reports `DependentClasses: ["Helper"]`, `TypeHintDependencies: 1`, and `CouplingCount: 1`.

The nested-class resolution check should also be applied when collecting type-hint dependencies, otherwise local helper classes can still be counted as coupling through annotations.

StressTestor:
You're right again, thanks. The annotation path runs through `extractTypeAnnotationDependencies`, which wasn't applying the scope check.

I've pulled the scope-aware resolution into a small resolver that's now applied on every path that can bind a nested class: instantiation, attribute access, and type annotations (both variable annotations and method parameter/return annotations). So `item: Helper` in your example is excluded the same way `Helper()` already was, while a same-named class reached from another scope still counts.

Added a test for the annotation case (`LocalHelperNotCountedViaAnnotation`) alongside the other two. Pushed.

DaisukeYoda:
@StressTestor 
I think the remaining issue is that method body scope and method annotation scope are being treated the same.

For normal method body references, a class-body nested class is not visible:

```python
class Outer:
    class Helper:
        pass

    def make(self):
        return Helper()  # does not resolve to Outer.Helper
```

But annotations on the function definition are evaluated while the class body is being executed, so this does resolve to the class-body nested class:

```python
class Outer:
    class Helper:
        pass

    def make(self) -> Helper:
        ...
```

So even though the annotation node is under `FunctionDef` in the AST, its name-resolution context should be the class body, not the method body. Right now `enclosingScope` treats it as the method scope, so `Helper` is still counted as external coupling.

The resolver likely needs to distinguish references inside a method body from references in method signature annotations.

StressTestor:
you're right, that's the gap. fixed it.

method signature annotations (params and return) now resolve in the method's *defining* scope, not the method body. so a class-body nested class is visible to them and gets excluded, while a method-body reference still can't see it and stays counted:

```python
class Outer:
    class Helper:
        pass

    def make(self) -> Helper:   # excluded - annotation evaluates in the class body
        ...

    def use(self):
        return Helper()         # still counts - method body can't see Outer.Helper
```

mechanically: the type-hint walk now carries an evaluation scope. method-signature annotations get the method's enclosing scope; class-body and method-body annotations keep resolving at their own lexical position as before, and the instantiation/attribute path is unchanged. i also folded the body and signature resolution onto a shared `resolvesInScope` so they can't drift apart again.

tests cover the bare (`-> Helper`), generic (`-> list[Helper]`), and union (`Helper | None`) signature forms, plus a method-body complement that pins it still counts.

one honest limitation i didn't change here: resolution is structural, not execution-order aware. a nested class defined textually after the method but referenced in that method's signature is still treated as resolvable, where real CPython would raise NameError at class-body execution time. it lines up with how the rest of the analyzer resolves names, so it felt out of scope, but happy to tighten it to definition order if you'd prefer.

DaisukeYoda (approved):
Thanks for coming back! LGTM!

PR #606, opened 2026-06-26
docs: clarify architecture defaults

### LINKED ISSUE #524: Misleading 'disabled by default' comments for architecture analysis (clarify SOT)
## Summary

The default-enabled state of **architecture analysis** is defined in multiple places with **conflicting values**, and the struct-level defaults carry comments that contradict the actual runtime behavior. This is a documentation/consistency cleanup — no behavior change is required.

## Background

For the `analyze` command, whether architecture analysis runs is gated **solely** by `AnalyzeExecutionConfig.SystemAnalyzeArchitecture`:

- **No config file** → `defaultAnalyzeExecutionConfig()` sets it to `true` (`service/analyze_config_loader.go:76-78`).
- **Config file present** → only an explicit `[architecture] enabled = <bool>` overrides it (`service/analyze_config_loader.go:145-147`). Omitting the key leaves architecture **enabled**.

So architecture analysis is **enabled by default**.

However, two struct defaults say the opposite, with misleading comments:

```go
// internal/config/config.go:210
Enabled: false, // Disabled by default - opt-in feature
```
```go
// internal/config/pyscn_config.go:387
ArchitectureEnabled: domain.BoolPtr(false), // Disabled by default - opt-in
```

These `false` values feed `config.Config.Architecture.Enabled`, but that field is **only ever written, never read** to gate analysis (verified — the only two references are assignments):

- `internal/config/config.go:461` (write)
- `service/config_loader.go:275` (write)

The real gate is `SystemAnalyzeArchitecture` (default `true`). So the `// Disabled by default - opt-in feature` comments are inaccurate, and `Architecture.Enabled` is effectively a dead field for gating. A newcomer reading these comments would reasonably (and wrongly) conclude architecture is off by default.

## Why this matters

- The comments actively mislead. Someone could later add code that reads `cfg.Architecture.Enabled` assuming `false`/opt-in, silently breaking the actual default-on behavior.
- There is no single source of truth for "is architecture enabled by default".

## Scope (pick the minimal safe fix)

**Option A — comment-only fix (recommended for a first PR):**
Update the two comments at `config.go:210` and `pyscn_config.go:387` to reflect reality, e.g.:
> The effective default for the `analyze` command comes from `defaultAnalyzeExecutionConfig()` (`SystemAnalyzeArchitecture: true`); this struct field is not read to gate analysis.

**Option B — remove the dead field (optional follow-up):**
If investigation confirms `config.Config.Architecture.Enabled` is never read anywhere for gating or reporting, consider removing it (and the writes at `config.go:461` / `config_loader.go:275`) to leave a single source of truth. Verify report output / serialization does not depend on it first.

## Acceptance criteria

- [ ] Comments no longer claim architecture is "disabled by default / opt-in" when it is enabled by default.
- [ ] (If Option B) `Architecture.Enabled` removed only after confirming no reader depends on it; `make test` and `make lint` pass.
- [ ] No behavior change to `pyscn analyze` (architecture still runs by default).

## Pointers

- Effective default (true): `service/analyze_config_loader.go:76-78`
- Override gate: `service/analyze_config_loader.go:142-147`
- Misleading defaults/comments: `internal/config/config.go:210`, `internal/config/pyscn_config.go:387`
- Dead-field writes: `internal/config/config.go:461`, `service/config_loader.go:275`

## Context

Surfaced while reviewing #335 (architecture style presets). The new `style` feature is unaffected — it works because architecture runs by default.

sanmaxdev:
## Summary
Clarifies the two architecture config comments that still described architecture analysis as disabled by default, while the analyze command uses `SystemAnalyzeArchitecture` as the effective gate.

## Type of Change
- [ ] Bug fix (non-breaking change)
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues
Closes #524

## Changes
- Updated the `DefaultConfig()` architecture comment to point at the effective analyze gate.
- Updated the `DefaultPyscnConfig()` architecture comment with the same source-of-truth note.

## Testing
- [x] `go test ./internal/config`
- [x] `go vet ./internal/config`
- [x] `go test ./...`
- [x] `go build ./cmd/pyscn`
- [x] `git diff --check`

## Documentation
- [x] Updated relevant config comments


internal/config/config.go
```diff
@@ -235,7 +235,7 @@ func DefaultConfig() *Config {
 
 		// Architecture validation configuration
 		Architecture: ArchitectureConfig{
-			Enabled:                         false, // Disabled by default - opt-in feature
+			Enabled:                         false, // Config model default only; analyze uses SystemAnalyzeArchitecture as the effective gate.
```
DaisukeYoda: Extend the comment along the lines of issue #524 Option A, e.g. `// Config model default only; not read to gate analysis. Analyze effective default: SystemAnalyzeArchitecture=true in defaultAnalyzeExecutionConfig().`

internal/config/pyscn_config.go
```diff
@@ -399,7 +399,7 @@ func DefaultPyscnConfig() *PyscnConfig {
 		LcomMediumThreshold: domain.DefaultLCOMMediumThreshold,
 
 		// Architecture defaults (from [architecture] section)
-		ArchitectureEnabled:                         domain.BoolPtr(false), // Disabled by default - opt-in
+		ArchitectureEnabled:                         domain.BoolPtr(false), // Config model default only; analyze uses SystemAnalyzeArchitecture as the effective gate.
```
DaisukeYoda: Add a short file/function hint (e.g. `defaultAnalyzeExecutionConfig` in `service/analyze_config_loader.go`) so the symbol is traceable from the config package.

DaisukeYoda (changes requested):
Thanks! Please take a look at my review comments.

internal/config/config.go
sanmaxdev: Updated per your suggestion. The comment now reads: `// Config model default only; not read to gate analysis. Analyze effective default: SystemAnalyzeArchitecture=true in defaultAnalyzeExecutionConfig().`

internal/config/pyscn_config.go
sanmaxdev: Added the file hint. The comment now reads: `// Config model default only; not read to gate analysis. Analyze effective default: SystemAnalyzeArchitecture=true in defaultAnalyzeExecutionConfig() (see service/analyze_config_loader.go).`

DaisukeYoda (approved):
LGTM!

PR #615, opened 2026-07-01
fix: drop clone groups without backing pairs

### LINKED ISSUE #539: [BUG][auto] type4-zero-similarity-group: clone group emitted with similarity=0.0 and no backing clone pair
# Bug description

`pyscn analyze` emitted a type-4 clone group with `similarity: 0` — a self-contradictory value for a clone group (members of a clone group are by definition similar, and the configured type-4 threshold is far above 0). The group also has **no backing entry in `clone_pairs`**: every other group in the run corresponds to pairs with `similarity >= 0.85`, but no pair exists between these two fragments at all. Whatever path emitted this group either skipped the similarity computation (field left at zero value) or bypassed the pair-threshold check.

# Repro steps

```bash
git clone https://github.com/tlsfuzzer/tlslite-ng && cd tlslite-ng
git checkout 02d1506badb16473faf50ebc3a413c6d789fe31f
jq '.clone.clone_groups[] | select(.similarity < 0.5)' .pyscn/reports/analyze_*.json
```

The flagged source (`tlslite/tlsconnection.py:3931-3944`) is two adjacent, structurally near-identical guard blocks inside a generator method:

```python
                if clientHello.srp_username:
                    if not session.srpUsername or \
                            clientHello.srp_username != \
                            bytearray(session.srpUsername, "utf-8"):
                        for result in self._sendError(
                                AlertDescription.handshake_failure):
                            yield result
                if clientHello.server_name:
                    if not session.serverName or \
                            clientHello.server_name != \
                            bytearray(session.serverName, "utf-8"):
                        for result in self._sendError(
                                AlertDescription.handshake_failure):
                            yield result
```

A minimal synthetic repro is likely two adjacent `if x: if not a or x != f(a): for r in g(): yield r` blocks in one generator method, but the trigger condition (group emitted without a qualifying pair) suggests the bug is in group construction/aggregation rather than in any particular source shape.

# Expected behavior

A clone group's `similarity` should reflect the (aggregate of) member-pair similarity and always be at or above the detection threshold; a group with no qualifying pair should not be emitted at all.

# Actual output

```json
{
  "id": 49,
  "type": 4,
  "similarity": 0,
  "size": 2,
  "clones": [
    {"id": 1413, "type": 4, "location": {"file_path": "tlslite/tlsconnection.py", "start_line": 3931, "end_line": 3937, "start_col": 16, "end_col": 40}, "size": 26, "line_count": 7},
    {"id": 1415, "type": 4, "location": {"file_path": "tlslite/tlsconnection.py", "start_line": 3938, "end_line": 3944, "start_col": 16, "end_col": 40}, "size": 26, "line_count": 7}
  ]
}
```

Cross-checks on the same report:
- `jq '[.clone.clone_groups[] | select(.similarity < 0.85)]'` → only this group.
- No `clone_pairs` entry references fragments at lines 3931/3938 of `tlslite/tlsconnection.py` (locations checked directly since pair fragment ids are all 0 — see #488).

# Environment

- pyscn `v1.23.2-5-g29e91cd`

# Priority rationale

P2 — visible but bounded inaccuracy: a single internally inconsistent group-level field; the cited fragments are real and adjacent, so the gist of the finding is right.

---

Found via the FP-audit skill in repo `tlsfuzzer/tlslite-ng@02d1506badb16473faf50ebc3a413c6d789fe31f`.

sanmaxdev:
## Summary
- Drops clone groups whose remaining members no longer have any backing clone pair after group deduplication.
- Refreshes final group metadata from surviving pairs so unbacked groups do not report similarity 0.

## Type of Change
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues
Fixes #539

## Changes
- Recompute final clone group metadata after group deduplication.
- Filter clone groups that have no backing pair among remaining members.
- Add a regression test for unbacked clone groups.

## Testing
- [x] `go test ./internal/analyzer -run 'TestFilterCloneGroupsWithoutBackingPairs|TestGroupClonesWithStrategy_FiltersSuppressedClonePairs|TestCloneDetector_PairsPromotedToGroups' -v`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build -v ./cmd/pyscn`
- [x] `go run ./cmd/pyscn analyze --json --no-open /tmp/tlslite-ng`
- [x] `git diff --check`

## Documentation
- [ ] Updated godoc comments

### DIFF
--- a/internal/analyzer/clone_detector.go
+++ b/internal/analyzer/clone_detector.go
@@ -1194,7 +1194,7 @@ func (cd *CloneDetector) groupClonesWithStrategy(strategy GroupingStrategy) {
 	}
 	memberResult := dedupeStrictSubsetGroupMembers(strategy.GroupClones(cd.clonePairs), cd.clonePairs)
 	groupResult := dedupeCoveredGroups(memberResult.groups)
-	cd.cloneGroups = groupResult.groups
+	cd.cloneGroups = filterCloneGroupsWithoutBackingPairs(groupResult.groups, cd.clonePairs)
 	for fragment := range memberResult.suppressed {
 		groupResult.suppressed[fragment] = struct{}{}
 	}
--- a/internal/analyzer/group_dedup.go
+++ b/internal/analyzer/group_dedup.go
@@ -293,6 +293,25 @@ func refreshGroupMetadata(group *CloneGroup, similarities map[string]float64, cl
 	group.CloneType = majorityCloneType(cloneTypes, similarities, group.Fragments)
 }
 
+func filterCloneGroupsWithoutBackingPairs(groups []*CloneGroup, pairs []*ClonePair) []*CloneGroup {
+	if len(groups) == 0 {
+		return groups
+	}
+	similarities, cloneTypes := clonePairMetadata(pairs)
+	out := make([]*CloneGroup, 0, len(groups))
+	for _, group := range groups {
+		if group == nil || len(group.Fragments) < 2 {
+			continue
+		}
+		refreshGroupMetadata(group, similarities, cloneTypes)
+		if group.Similarity <= 0 {
+			continue
+		}
+		out = append(out, group)
+	}
+	return out
+}
+
 func filterClonePairsWithSuppressedMembers(pairs []*ClonePair, suppressed map[*CodeFragment]struct{}) []*ClonePair {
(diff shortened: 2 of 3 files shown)


DaisukeYoda (approved):
LGTM!

PR #632, opened 2026-07-08
fix: deduplicate clone pairs before reporting

### LINKED ISSUE #626: [BUG][auto] duplicate-clone-pair: clone pair appears multiple times in same clone group
## Bug Description
When clone detection groups semantically similar code fragments into `clone_groups`, a single clone pair can appear **multiple times** within the same group's pair list. In the `boxed/mutmut` repo, the pair `type_checking.py:44-58 ↔ type_checking.py:61-76` appears twice in group 1's pair collection.

## Repro Steps
Run pyscn clone detection on a repo that produces multi-clone groups (Type-4 semantic clones). Inspect the `clone_pairs` for each group — expect all pairs within a group to be unique.

## Expected Behavior
Each unique clone pair should appear exactly once per group. A pair `(A,B)` should not be duplicated as a second `(A,B)` entry in the same group.

## Actual Output
```json
{
  "clone_groups": [
    {
      "id": 1,
      "clone_pairs": [
        {"clone1": "type_checking.py:44-58", "clone2": "type_checking.py:61-76"},
        {"clone1": "type_checking.py:61-76", "clone2": "type_checking.py:79-93"},
        {"clone1": "type_checking.py:79-93", "clone2": "type_checking.py:96-112"},
        {"clone1": "type_checking.py:44-58", "clone2": "type_checking.py:61-76"}  // DUPLICATE
      ]
    }
  ]
}
```

The pair `(44-58, 61-76)` is listed twice.

## pyscn Version
v1.26.1

## Priority Rationale (P3)
P3 — cosmetic / low-impact visibility issue. The extra pair doesn't change the clone group analysis or metrics; it's a bounded off-by-one-like error in pair deduplication during group construction.

## Good First Issue
Yes — this is likely a missing deduplication guard in the clone group pair-collection loop, a bounded fix at one site.

Found via the FP-audit skill in repo `boxed/mutmut@523a19c`.

sanmaxdev:
## Summary
Deduplicate clone pairs by canonical fragment location before sorting and limiting results.

## Type of Change
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues
Closes #626

## Changes
- Keep one reported clone pair per unordered fragment-location pair.
- Prefer the higher-similarity pair when duplicates are produced.
- Add regression coverage for duplicate and reversed clone-pair entries.

## Testing
- [x] `go test ./internal/analyzer -run 'TestCloneDetector_limitAndSortClonePairs_DeduplicatesByLocation|TestCloneDetector_GetStatistics'`
- [x] `go test ./internal/analyzer`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `go build -v ./cmd/pyscn`

## Documentation
- [ ] Updated godoc comments

### DIFF
--- a/internal/analyzer/clone_detector.go
+++ b/internal/analyzer/clone_detector.go
@@ -1258,8 +1258,10 @@ func (cd *CloneDetector) tryCreateClonePair(i, j int, minSimilarity float64) *Cl
 	return nil
 }
 
-// limitAndSortClonePairs ensures final results are sorted and limited
+// limitAndSortClonePairs ensures final results are unique, sorted, and limited.
 func (cd *CloneDetector) limitAndSortClonePairs(maxPairs int) {
+	cd.clonePairs = dedupeClonePairs(cd.clonePairs)
+
 	// Sort clone pairs by similarity (descending)
 	sort.Slice(cd.clonePairs, func(i, j int) bool {
 		return cd.clonePairs[i].Similarity > cd.clonePairs[j].Similarity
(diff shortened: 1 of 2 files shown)


DaisukeYoda:
Thanks for the fix, but closing without merging — the underlying issue (#626) doesn't reproduce; see the explanation there.

The dedup logic in `dedupeClonePairs` guards against a scenario that can't actually occur given how `cd.clonePairs` is built (verified both by code inspection and an empirical repro test with mutually-similar fragments, which produced zero duplicates via both the exhaustive and LSH detection paths). The new test also only exercises the helper directly by hand-injecting a duplicate pair into the private `clonePairs` field, rather than going through the real detection pipeline, so it doesn't demonstrate the bug happening in practice. We'd rather not carry the extra code path for a case that can't happen today.

PR #634, opened 2026-07-08
fix(cbo): exclude annotation-only deps under PEP 563 future annotations

### LINKED ISSUE #628: [BUG][auto] annotation-only-deps: CBO metric counts future-annotations type hints as coupling
**Bug**: pyscn's CBO metric counts class-level attribute type annotations as `ImportDependencies` at full weight, even when `from __future__ import annotations` makes them strings at runtime with zero import cost.

**Repro** (minimal Python):
```python
# repro/tokens_example.py
from __future__ import annotations
import ast

class TOKENS:
    """No methods, just type-annotated constants."""
    ASSERT: type[ast.Assert]
    ATTRIBUTE: type[ast.Attribute]
    CALL: type[ast.Call]
    RETURN: type[ast.Return]
```

Run: `pyscn analyze --select cbo repro/`

**Expected**: CBO ≈ 0 (no runtime coupling; all `ast.*` annotations are strings under PEP 563).

**Actual**: CBO = 4 (12 in the real-world case with 12 ast types). pyscn counts every annotation type reference as an `ImportDependency`, inflating CBO for simple enum-like constant classes that have zero behavioral coupling.

**pyscn version**: v1.26.1

**Priority rationale (P1)**: A user seeing CBO=12 on a 12-line constant class with zero methods would be substantially misled into thinking the class is architecturally coupled when it has no behavioral dependencies at all.

**Found via**: FP-audit skill in repo `life4/deal@751c0a7e`.

vaishnavidesai09:
## Summary

CBO counted every type-annotation reference as coupling, even under `from __future__ import annotations` (PEP 563), where annotations are stored as strings and never evaluated at runtime. A 12-line constant class annotated purely with `ast.*` types reported CBO=12 despite having zero actual behavioral coupling (see #628).

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues

Fixes #628

## Changes

- Added a `futureAnnotations` flag to `CBOAnalyzer`, detected once per file via a new `hasFutureAnnotations()`.
- `from __future__ import annotations` is parsed by tree-sitter as its own `future_import_statement` grammar node, not the general `import_from_statement` that pyscn's import collector already handles, so detection inspects the raw node's children for an `annotations` token rather than relying on the usual `Module`/`Names` fields.
- `addDependency()` now skips adding a dependency edge for `dependencyKindTypeHint` when `futureAnnotations` is active. Inheritance, instantiation, and attribute-access edges are unaffected, since those still execute at runtime under PEP 563.
- Updated `TestCBOAnalyzer_DependencyIdentityContract`: its fixture already had `from __future__ import annotations` and was asserting the old (technically incorrect under PEP 563) CBO=1 for annotation-only references to a local class. Now asserts CBO=0, with a comment explaining why.

## Testing

- [x] `go test ./internal/analyzer/...` passes locally (all existing tests + 3 new regression tests)
- [x] `make lint` — not yet run, will confirm
- [x] Manual testing completed

Added 3 regression tests:
- `TestCBOAnalyzer_FutureAnnotationsExcludeAnnotationOnlyDependencies` — reproduces the issue's exact repro, confirms CBO 4 → 0
- `TestCBOAnalyzer_FutureAnnotationsStillCountRuntimeUsage` — confirms instantiation/call usage still counts even with the future import present, only pure-annotation references are excluded
- `TestCBOAnalyzer_WithoutFutureAnnotationsTypeHintsStillCount` — confirms pre-#628 behavior is unchanged when the future import is absent

Also manually verified against the CLI using the issue's repro file plus a mixed file (one annotation-only class, one class with real runtime coupling) — confirmed the annotation-only class drops to CBO=0 while the class with actual `ast.Call()` usage still reports CBO=1.

## Documentation

- [x] Updated godoc comments (`hasFutureAnnotations`, `importsAnnotationsName`, the `addDependency` skip, and the `futureAnnotations` field all documented inline)
- [ ] Updated README or other docs (not needed — internal analyzer behavior only)

## Additional Context

None of this affects `ImportDependencies` accounting for annotations that *aren't* future-only — e.g. a plain `x: ast.AST` without the future import still counts exactly as before. Only annotation-only references in files with `from __future__ import annotations` are affected.

### DIFF
--- a/internal/analyzer/cbo.go
+++ b/internal/analyzer/cbo.go
@@ -76,6 +76,13 @@ type CBOAnalyzer struct {
 	importedNames    map[string]string         // alias -> module.name mapping
 	namespaceAliases map[string]string         // module.name -> alias mapping (only for alias imports)
 	regexCache       map[string]*regexp.Regexp // pattern -> compiled regex cache
+
+	// futureAnnotations is true when the file being analyzed has
+	// `from __future__ import annotations` (PEP 563) at module scope. Under
+	// PEP 563 all annotations are evaluated as strings at runtime, so a class
+	// referenced only inside a type annotation is never actually imported or
+	// resolved and imposes zero coupling cost. File-local, reset per AST. See #628.
+	futureAnnotations bool
 }
 
 type cboDependencyKind int
@@ -149,6 +156,7 @@ func (a *CBOAnalyzer) AnalyzeClasses(ast *parser.Node, filePath string) ([]*CBOR
 	// cannot change dependency identity in the next file.
 	a.importedNames = imports.importedNames
 	a.namespaceAliases = imports.namespaceAliases
+	a.futureAnnotations = a.hasFutureAnnotations(ast)
 
 	var results []*CBOResult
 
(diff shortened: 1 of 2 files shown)


vaishnavidesai09:
@DaisukeYoda review this pr!

DaisukeYoda (approved):
LGTM! Thanks

PR #635, opened 2026-07-08
feat(complexity): report pre-filter function count in summary (#597)

### LINKED ISSUE #597: [ENH][auto] function-coverage-disclosure: report all functions in summary or disclose min_complexity filtering
**Track**: feature-gap

## Summary
pyscn's `total_functions` summary field reports only post-filter functions (McCabe complexity ≥5), misleading users into thinking it covers the full codebase. The filtering threshold is effectively hidden — the `--min-complexity` flag defaults to 5 at the CLI level (`cmd/pyscn/analyze.go:137`), but `domain/defaults.go:116` sets `DefaultComplexityMinFilter = 1`. The resulting `total_functions: 75` (when 241 actual functions exist) inflates the average complexity and gives a false sense of thoroughness.

**Rationale (user value)**: Users reading a summary that says `total_functions: 75` reasonably assume pyscn analyzed 75 functions. They don't know 166 functions were silently dropped because they had trivial cyclomatic complexity. The health score (e.g. 66/C) is based on only 31% of the codebase's functions. This erodes trust.

## Evidence

**Repo**: pg-spot-ops/pg-spot-operator @ 41def6000cc2fdb0f2007b9954ddb40177f9ec6f

- `pg_spot_operator/util.py`: 27 `def` statements, pyscn reports **2** functions in `Functions[]`
- `pg_spot_operator/cloud_impl/aws_cache.py`: 10 `def` statements, pyscn reports **0** functions in `Functions[]`
- Total: ~241 non-test function definitions across 27 analyzed files, only 75 reported (31%)
- pyscn output: `summary.total_functions: 75`, `summary.average_complexity: 9.04`, complexity range in `Functions[]` is 5–41 — no function with C<5 appears

Synthetic repro:
```python
# trivial_function.py
def add(a, b):   # C=1 — dropped by min_complexity=5 filter
    return a + b

def factorial(n):  # C=2 — dropped
    if n <= 1:
        return 1
    return n * factorial(n - 1)

def complex_func(x):  # C=8 — reported
    result = 0
    for i in range(x):
        if i % 2 == 0:
            result += i
        elif i % 3 == 0:
            result -= i
        else:
            result += 1
    return result
```
After `pyscn analyze --json trivial_function.py`, `total_functions: 1` despite the file having 3 functions.

## Root-cause hypothesis
`filterFunctions` at `service/complexity_service.go:283` drops all functions with `Complexity < MinComplexity`. `generateSummary` at line 395 sets `TotalFunctions = len(post-filter slice)`. The CLI flag `--min-complexity` defaults to 5 (`cmd/pyscn/analyze.go:137`), which overrides the domain default of 1 (`domain/defaults.go:116`). There is no `functions_analyzed` or `functions_total` field.

## Proposed change
**Option A (preferred)**: Keep all functions in `Functions[]` regardless of complexity, and let the `min_complexity` filter control only the risk-level classification and highlighting in the summary. The `total_functions` field then reflects actual coverage.

**Option B (non-breaking)**: Add `functions_analyzed` (post-filter count) and `functions_filtered` (dropped count) to the summary alongside the existing `total_functions`. Document the `min_complexity` default behavior explicitly in the summary.

## Impact / Effort / Confidence
- **Impact**: medium
- **Effort**: medium
- **Confidence**: high

## Verification
Survived refutation — the evidence holds: util.py has 27 defs vs 2 reported, aws_cache.py has 10 defs vs 0. The `--min-complexity` default of 5 is documented in `--help` but the filtering effect is invisible in the summary. The field naming is misleading.

## pyscn version
v1.24.3-14-g67b243f

Found via the pyscn-enhancement-audit skill in repo `pg-spot-ops/pg-spot-operator@41def6000cc2fdb0f2007b9954ddb40177f9ec6f`.

DaisukeYoda:
## Summary

Add `FunctionsTotal` field to `ComplexitySummary` so consumers can see how many functions were parsed before `min_complexity` filtering dropped them. Previously `total_functions` only reflected the post-filter count, misleading users into thinking pyscn analyzed fewer functions than it actually parsed (e.g., 75 reported vs 241 actual in `pg-spot-ops/pg-spot-operator`).

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Related Issues

Closes #597

## Changes

- Added `FunctionsTotal` field to `domain.ComplexitySummary` — holds the pre-filter count (all parsed functions before `min_complexity` filtering)
- Updated `generateSummary` to accept and propagate the pre-filter count
- Updated `Analyze` and `AnalyzeSnapshot` callers to pass `len(allFunctions)` before filtering
- Exposed `functions_total` in JSON/YAML summary output and in text formatter

## Testing

- [x] `make test` passes (pre-existing integration test failures in `TestArchitecture_*Preset` are unrelated)
- [x] `make lint` passes (0 issues)
- [x] Added regression test: `TestComplexityService_GenerateSummary/FunctionsTotal_tracks_pre-filter_count_when_min_complexity_drops_functions`
- [x] `go vet ./...` clean
- [x] `go build ./cmd/pyscn` succeeds

## Documentation

- [ ] Not applicable — `functions_total` is self-documenting in JSON/YAML output

---
Generated by the pyscn-issue-fix skill from issue #597. Review before marking ready.


DaisukeYoda:
Rename before merge: `TotalFunctions` / `FunctionsTotal` (and the text/JSON labels) are nearly identical, so users still cannot tell pre- vs post-filter. Prefer clearer names such as `functions_parsed` or an explicit filtered count. Also, unified analyze summary, analyze text/HTML, and MCP still only show the post-filter total—either propagate the disclosure there or call out that limited scope in the PR. A single “reported vs analyzed” line would be clearer than two near-identical summary fields.

PR #638, opened 2026-07-09
fix: ignore module constants in CBO dependencies

### LINKED ISSUE #636: [BUG][auto] cbo-counts-module-constants-as-class-dependencies: re.DOTALL/MULTILINE and subprocess.PIPE counted as CBO dependent classes
### Bug Description

pyscn's CBO analyzer counts `re.DOTALL` and `re.MULTILINE` as class dependencies, inflating coupling scores. These are `int`-typed module-level flag constants (values 16 and 8), not classes. This causes the `FileContent` class in `tfeldmann/organize` to be tagged **high-risk** with CBO=8, when the real class-dependency count is 6 (medium-risk).

### Repro Steps

Create a minimal Python file:

```python
# coupling_artifact.py
import re
from pathlib import Path

class UsesRegexFlags:
    def search_content(self, path: Path, pattern: str) -> bool:
        return re.compile(pattern, re.DOTALL | re.MULTILINE).search(path.read_text()) is not None
```

Run `pyscn analyze --json --no-open .` and inspect `.cbo.Classes` for `UsesRegexFlags`:

```bash
jq '.cbo.Classes[] | select(.Name == "UsesRegexFlags") | .Metrics.DependentClasses'
```

### Expected Behavior

`DependentClasses` should only list actual classes: `["Path"]`. Module-level flag constants (`re.DOTALL`, `re.MULTILINE`) and other non-class symbols should not appear.

### Actual Output

```json
["Path", "re.DOTALL", "re.MULTILINE", "re.Match"]
```

From `tfeldmann/organize` (FileContent class):
```json
["ConfigDict", "FilterConfig", "Output", "Path", "Resource", "re.DOTALL", "re.MULTILINE", "re.Match"]
```

3 out of 8 listed "dependencies" are non-class symbols. This inflates coupling from 6 (medium) to 8 (high-risk).

### Priority Rationale (P1)

Substantially misleading output a user would act on: a simple utility class using regex compilation flags gets incorrectly flagged as high-coupling. Affects any codebase that uses `re.compile()` with flags or `subprocess` constants (`subprocess.PIPE`, `subprocess.STDOUT`).

### Environment
- **pyscn**: v1.26.1-12-gdc83de5
- **Repo**: tfeldmann/organize@36a54572488d89dc9279d79848ecc067b632f1a5
- Found via FP-audit skill

sanmaxdev:
## Summary
- Ignore imported module-level constants such as `re.DOTALL` and `subprocess.PIPE` when calculating CBO dependencies.
- Keep class-like imported attributes counted while preventing constants from inflating coupling scores.

## Type of Change
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues
Closes #636

## Changes
- Added a guard for uppercase imported module constants in CBO dependency resolution.
- Added regression coverage for regex and subprocess constants.

## Testing
- [x] `go test ./internal/analyzer -run 'TestCBOAnalyzer_IgnoresImportedModuleConstants|TestCBOAnalyzer_AnalyzeClasses'`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./cmd/pyscn`
- [x] `git diff --check`

## Documentation
- [ ] Updated godoc comments


DaisukeYoda:
Thanks for the focused fix. The guard sits in the right place next to the imported-class binding preference, and the regression test covers the reported re/subprocess constant false positives well.

I would not merge this as-is. Treating any all-uppercase leaf as a non-class is too broad: it correctly drops constants like re.DOTALL and subprocess.PIPE, but it also drops real acronym-style classes such as uuid.UUID, httpx.URL, and models.Q. Those still count on main and disappear with this change.

Please narrow the constant filter to non-call attribute reads so call/instantiation targets still count, and add coverage that keeps uuid.UUID-style coupling intact.

sanmaxdev:
Pushed the requested narrowing: the uppercase-leaf filter now applies only to non-call attribute reads, while constructor calls such as `uuid.UUID()` and `httpx.URL()` still count.

Verification:
- `go test ./internal/analyzer -run 'TestCBOAnalyzer_(IgnoresImportedModuleConstants|KeepsUppercaseConstructorDependencies)' -count=1`
- `go test ./internal/analyzer -count=1`
- `go vet ./...`
- `go test ./...`
- `go build ./cmd/pyscn`
- `make lint`
- `git diff --check`

DaisukeYoda (approved):
Thanks for the follow-up. LGTM!

PR #649, opened 2026-07-13
refactor!: unify request bool override semantics

### LINKED ISSUE #641: Unify overridable plain bool request fields to *bool (nil = unset)
## Background

PR #640 unified all `MergeConfig` implementations on the zero-value sentinel convention (0 / "" / nil = unset) and converted `domain.CloneRequest`'s overridable bools to `*bool`. However, other request structs still carry plain `bool` fields, which cannot represent the three states a merge needs: **unset / explicit true / explicit false**.

## Current compromises

Plain bool fields force one of two broken merge patterns:

**Only-true-wins** (`service/config_loader.go` `ShowDetails`):
```go
if override.ShowDetails {
    merged.ShowDetails = override.ShowDetails
}
```
→ A config file with `show_details = true` cannot be turned off from the CLI; explicit `false` is indistinguishable from "not set".

**Always-override** (`service/config_loader.go` `Recursive`):
```go
merged.Recursive = override.Recursive
```
→ If the caller doesn't populate the field, the config file value is silently clobbered with `false`. Relies on every caller pre-filling correct values.

## Proposal

Convert the remaining overridable plain bool fields to `*bool` (nil = unset) and merge them with `config.MergePtr`, matching the pattern already used by `ComplexityRequest.Enabled`/`ReportUnchanged`, `DeadCodeRequest.DetectAfter*`, `CBORequest.ShowZeros`, and (since #640) all of `CloneRequest`'s bools.

Candidates (non-exhaustive; audit each `MergeConfig` for the two patterns above):
- `ComplexityRequest`: `ShowDetails`, `Recursive`
- `DeadCodeRequest`: `Recursive`, `ShowDetails` (if present)
- `CBORequest`: `ShowDetails`, `Recursive` (plain remnants)
- `LCOMRequest` / `SystemAnalysisRequest` / `DIAntipatternRequest` / `MockDataRequest`: same audit
- `NoOpen` across requests (currently always-override; decide whether it should participate in merge at all)

## Notes / pitfalls learned in #640

- Go `text/template`: a non-nil `*bool` is truthy even when it points to `false`. Any template referencing these fields must go through a resolving method (e.g. `ShouldShowContent()`), not the raw field.
- Callers that pre-fill requests with defaults must become sparse at the same time, or the non-nil pointer will override config values (the #640 review found several of these).
- Validation that requires complete requests must run after `MergeConfig`, with fail-fast checks only for caller-only inputs (paths, output writer).

## Out of scope

- `EnableDFA` on `CloneRequest` is intentionally not merged today (base always wins); changing that is a separate behavioral decision.

Follow-up to #640 / #553.

yhay81:
## What

- change the remaining overridable request booleans to `*bool` (`nil` = unset)
- let explicit `true` and `false` override configuration without omitted values clobbering it
- make `pyscn check` construct sparse requests and propagate the selected `ConfigPath` through the same merge path
- make downstream services honor the effective merged options, so configuration changes behavior rather than metadata only
- keep former boolean fields concrete in complexity and dead-code response metadata
- keep the app layer independent of `internal/config`

Affected boolean fields: `ComplexityRequest.ShowDetails/Recursive`, `DeadCodeRequest.Recursive`, `CBORequest.ShowDetails`, `LCOMRequest.ShowDetails`, and `MockDataRequest.Recursive`.

## Why the review follow-up is part of the bool/config change

The pointer conversion fixes how an override is represented, but the contract is only correct if the full path preserves that distinction:

`CLI selection -> config loading -> request merge -> validation -> service behavior -> response metadata`

The review exposed a break in that path: `checkMockdata` left `ConfigPath` empty, so `pyscn check --select mockdata --config ...` could load an auto-discovered config instead of the explicitly selected one. The follow-up in `de5c556` therefore passes `ConfigPath` and adds a regression with conflicting explicit and auto-discovered configs.

The same audit found adjacent cases where the effective request was not observable end to end:

- `check` handlers populated defaults too early, which could turn an omitted command option into an override and clobber configuration-derived values; they now pass sparse requests
- `Execute` and `AnalyzeAndReturn` previously validated only the sparse caller request before configuration was merged; they now also validate the resolved request
- dead-code detection booleans were merged but their corresponding findings were not filtered, so an explicit/configured `false` had no behavioral effect
- mock-data `keywords`, `domains`, and documented `ignore_patterns` were loaded into the request but were not consistently consumed by the service, making the selected configuration appear ineffective

These changes do not add a new analysis category. They make the existing request fields and the `nil = unset / non-nil = explicit` semantics observable through the affected `pyscn check` command, use-case merge, and service filtering paths, while preserving the existing concrete boolean response metadata.

## Compatibility and scope

This is an intentional source-breaking Go API change requested and approved in #641, following the `CloneRequest` precedent in #640. Callers must use `domain.BoolPtr(...)` or leave these fields `nil`. Parallel compatibility fields were avoided because they would create two sources of truth.

`NewMockDataServiceWithConfig` keeps its constructor-supplied detector configuration behavior. The request-driven path consumes the merged request configuration. No new detector category is introduced.

`NoOpen` remains a plain caller-only bool; `CloneRequest.EnableDFA` remains out of scope.

## Integration

Rebased onto #656. Its MCP `recursive` override now propagates through the converted pointer fields, so file selection and response metadata remain consistent.

## Validation

- `go test -count=1 ./...`
- `go vet ./...`
- `go build ./...`
- `golangci-lint run ./...` — 0 issues
- all four GitHub CI checks pass
- explicit `true` / explicit `false` / omitted-value merge coverage
- explicit config vs conflicting auto-discovered config regression
- configured complexity exclusions and dead-code detection flags
- configured mock-data keywords, domains, and ignore patterns
- invalid mock-data ignore-pattern and constructor-configuration compatibility coverage
- concrete complexity and dead-code response metadata coverage

Closes #641


yhay81:
Could you please take another look?

The related PRs have now merged, and I rechecked this branch against the current `main` (including #654). It merges cleanly, and the full test, vet, build, and lint checks pass. This PR is ready for review.

DaisukeYoda (commented):
@yhay81 Sure! Sorry for the delay.  I have one thing to mention. Could you take a look at the below point?

In `checkMockdata`, `Recursive` is now `nil` so it can inherit configuration, but the request still does not set `ConfigPath: c.configFile`. As a result, `pyscn check --select mockdata --config ...` ignores the specified config and may use an auto-discovered config instead. Please pass `ConfigPath` here and add a regression test.

yhay81:
Thanks for catching this. Addressed in `de5c556`:

- `checkMockdata` now forwards `ConfigPath: c.configFile`, so an explicitly selected `--config` is used.
- Added `TestCheckMockdataUsesExplicitConfigPath`, with conflicting auto-discovered and explicitly selected configs to guard against this regression.
- Re-audited the sibling `check` paths and configuration-derived analysis settings, and added regression coverage for the adjacent cases found during that review.

The full local test/vet/build/lint suite passes, and all four GitHub CI checks are green. Could you please take another look?

DaisukeYoda (approved):
@yhay81 LGTM! ありがとうございました！

PR #650, opened 2026-07-13
docs(mcp): correct configuration discovery

yhay81:
## Summary

- replace the stale YAML/JSON/XDG configuration discovery description with the current TOML-only behavior
- document `PYSCN_CONFIG`, working-directory discovery, target resolution, and built-in fallback
- describe configuration consumption per tool instead of claiming one uniform startup snapshot

## Root cause

The README described removed discovery paths and oversimplified runtime loading. `analyze_code` resolves target configuration, several metric tools pass a configured path to request-time loaders, and `check_coupling` uses selected startup values plus domain defaults without constructing a CBO configuration loader.

## Impact

Documentation only. Runtime behavior is unchanged. Users are told to fix or unset an invalid `PYSCN_CONFIG`, and the tool-specific loading differences are now explicit.

## Validation

- `git diff --check`
- compared the text with `cmd/pyscn-mcp/main.go`, `internal/config`, and every MCP handler's configuration flow

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -156,10 +156,13 @@ home-directory-specific configuration locations are not searched.
 2. **Server working directory** - When `PYSCN_CONFIG` is unset, the server
    searches upward from its current working directory. A `.pyscn.toml` file
    takes priority over `pyproject.toml` with a `[tool.pyscn]` section.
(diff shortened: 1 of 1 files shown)


DaisukeYoda (approved):
LGTM! Thanks for the document update, that really helps!

PR #651, opened 2026-07-13
fix(mcp): publish implemented output controls

yhay81:
## Summary

- publish the already-implemented `output_mode` and `max_results` controls in MCP tool schemas
- publish `min_cbo` for `check_coupling`
- constrain output modes and non-negative result limits in JSON Schema
- expose count and threshold controls as integers, matching handler semantics
- document the controls, including the previously undocumented `check_cohesion` tool section

## Root cause

The handlers implemented these arguments, but the registered MCP schemas and README did not expose them. Schema-driven clients therefore could not discover supported functionality. Publishing count controls as generic numbers would also admit fractional values that handlers truncate.

## Impact

Default handler behavior is unchanged. Clients gain discoverable access to existing summary, detailed, and full output modes, while schema validation rejects fractional counts.

## Validation

- `go test -count=1 ./...`
- schema regression tests for every affected tool
- output-mode enum, non-negative minimum, and integer-type assertions

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -273,6 +273,8 @@ Field notes:
 - `min_complexity` (optional): Minimum complexity to report (default: `1`)
 - `max_complexity` (optional): Maximum allowed complexity, 0 = no limit (default: `0`)
 - `show_details` (optional): Include detailed metrics (default: `true`)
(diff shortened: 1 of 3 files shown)


DaisukeYoda (approved):
Thanks! Helpful

PR #654, opened 2026-07-13
fix(check): fail on malformed discovered .pyscn.toml

yhay81:
## Summary

- resolve the existing cwd-based `pyscn check` configuration once per invocation
- explicitly parse a resolved configuration before running analyzers
- fail the quality gate for a malformed discovered `.pyscn.toml` or malformed explicitly selected config
- keep unrelated malformed `pyproject.toml` files ignored

## Root cause and scope

Individual default configuration loaders discard load errors and fall back to hardcoded defaults. A malformed discovered `.pyscn.toml` could therefore turn a configured CI gate into a false success.

An invalid `pyproject.toml` cannot be reliably classified as containing a pyscn table by the TOML parser. The earlier text-scanning heuristic had both false positives and false negatives for quoted keys, conditional syntax damage, comments, and multiline strings, so it has been removed. This PR deliberately guarantees fail-fast behavior only where configuration intent is unambiguous: `.pyscn.toml` discovery and explicit config paths.

## Impact

Configuration precedence remains unchanged and discovery still starts from the current working directory. Unrelated broken `pyproject.toml` files are not newly made fatal.

## Validation

- `go test -count=1 ./...`
- malformed discovered `.pyscn.toml` fails
- unrelated malformed `pyproject.toml` remains ignored
- existing parent-directory configuration discovery remains green

### DIFF
--- a/app/analyze_usecase.go
+++ b/app/analyze_usecase.go
@@ -46,6 +46,12 @@ type AnalyzeUseCaseConfig struct {
 	Verbose    bool
 }
 
+// AnalyzeRequestOverrides contains request-scoped values that take precedence
+// over the resolved project configuration.
+type AnalyzeRequestOverrides struct {
+	Recursive *bool
+}
+
 // AnalyzeUseCase orchestrates comprehensive analysis
 type AnalyzeUseCase struct {
 	complexityUseCase *ComplexityUseCase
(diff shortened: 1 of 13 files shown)


DaisukeYoda (approved):
LGTM!

PR #656, opened 2026-07-13
fix(mcp): propagate analyze recursive override

yhay81:
> **Merge first.** Draft #649 depends on this PR; I will update and retest #649 afterward.

## Summary

- honor MCP `analyze_code.recursive`
- preserve project configuration when the argument is omitted
- propagate the effective value to file collection, child requests, and full-response metadata
- reject non-boolean values at the handler boundary

## Compatibility

Existing request types and `AnalyzeUseCaseConfig` remain unchanged. The additive override API keeps this PR v1-compatible; #649 handles the separately approved breaking migration.

## Validation

- `go test -count=1 ./...`
- explicit `true`/`false`, omitted configuration, full output, and invalid-type coverage

### DIFF
--- a/app/analyze_usecase.go
+++ b/app/analyze_usecase.go
@@ -46,6 +46,12 @@ type AnalyzeUseCaseConfig struct {
 	Verbose    bool
 }
 
+// AnalyzeRequestOverrides contains request-scoped values that take precedence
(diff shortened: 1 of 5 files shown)


DaisukeYoda (approved):
LGTM!

PR #657, opened 2026-07-14
fix: expose CBO value in class JSON output

### LINKED ISSUE #643: [BUG][auto] cbo-field-null-metrics-present: CouplingBetweenObjects field is always null while Metrics.CouplingCount has the real value
### Bug description

When pyscn v1.26.2 analyzes a Python repository with `--json`, every class in the `cbo.Classes` array has `CouplingBetweenObjects: null`, even though the underlying coupling data computed correctly and stored in `Metrics.CouplingCount` is accurate. The CBO summary (`.cbo.Summary`) correctly aggregates the real values, so the analysis engine works — but the per-class JSON field is unusable.

### Steps to reproduce

Run pyscn on any Python project with classes:

```bash
./pyscn analyze --json --no-open <any-repo>
```

In the output JSON, examine any entry under `.cbo.Classes[].CouplingBetweenObjects` — it is always `null` regardless of risk level.

### Expected behavior

`CouplingBetweenObjects` should contain the numeric CBO value for each class, matching or derived from `Metrics.CouplingCount`.

### Actual output (from `raw/analyze.json`)

```json
{
  "Name": "ResultBrowser",
  "CouplingBetweenObjects": null,
  "Metrics": {
    "CouplingCount": 11,
    ...
  },
  "RiskLevel": "high"
}
```

All 24 classes in the audit showed `CouplingBetweenObjects: null`, while `Metrics.CouplingCount` ranged from 1–11 (matching the summary's CBODistribution).

### pyscn version

v1.26.2 (commit 80ab47f)

### Priority rationale

**P2**: Visible but bounded inaccuracy — the value exists at `Metrics.CouplingCount` one level deeper, so the data is not lost. However, `CouplingBetweenObjects` at the top level is the canonical field name users and tooling expect, making the output misleading.

### Found via

FP-audit skill in repo `boxed/mutmut@523a19c07f4b920ebd19f306e2c414c9d0d28ca6`

sanmaxdev:
## Summary
- expose `CouplingBetweenObjects` as a numeric per-class JSON field
- populate it from the same analyzer result as `Metrics.CouplingCount`

## Type of Change

## Related Issues
Fixes #643

## Changes
- add the top-level CBO field to class results
- add conversion and JSON regression coverage

## Testing
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./cmd/pyscn`
- [x] JSON CLI reproduction confirms both CBO fields report `1`

## Documentation
- [x] No documentation changes needed

### DIFF
--- a/domain/cbo.go
+++ b/domain/cbo.go
@@ -66,7 +66,8 @@ type ClassCoupling struct {
 	EndLine   int
 
 	// CBO metrics
-	Metrics CBOMetrics
+	CouplingBetweenObjects int
+	Metrics                CBOMetrics
 
 	// Risk assessment
 	RiskLevel RiskLevel
--- a/service/cbo_formatter_test.go
+++ b/service/cbo_formatter_test.go
@@ -16,9 +16,10 @@ func createTestCBOResponse() *domain.CBOResponse {
 	return &domain.CBOResponse{
 		Classes: []domain.ClassCoupling{
 			{
-				Name:      "MyClass",
-				FilePath:  "test.py",
-				StartLine: 10,
+				Name:                   "MyClass",
+				FilePath:               "test.py",
+				StartLine:              10,
+				CouplingBetweenObjects: 5,
 				Metrics: domain.CBOMetrics{
 					CouplingCount:    5,
 					DependentClasses: []string{"ClassA", "ClassB", "ClassC"},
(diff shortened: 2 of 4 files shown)


DaisukeYoda (changes requested):
This change treats a non-existent schema field as a production bug. Class JSON already exposes the CBO value under Metrics.CouplingCount, and that matches the nested-metrics shape used by complexity and LCOM. The linked issue’s CouplingBetweenObjects: null shape does not appear in real analyze output for the cited case; the field has never been part of the domain model. Adding a top-level CouplingBetweenObjects that mirrors CouplingCount creates a second source of truth without fixing real output, and any construction path that only fills Metrics can desync the two values. Prefer closing the linked issue as invalid (or documenting CouplingCount as the canonical CBO value) rather than expanding the domain model for a hallucinated field.

sanmaxdev:
Thanks for catching this. `Metrics.CouplingCount` is the canonical CBO value, so I’m closing this change rather than adding a duplicate field.

Issue #20, opened 2025-08-07
[TASK] CFG Builder - Exception Handling (#3.5)


DaisukeYoda:
### Description
Implement exception handling in CFG builder (try/except/finally blocks).

### Implementation Tasks
- [ ] Implement try block with protected region
- [ ] Handle except clauses (single and multiple)
- [ ] Support except with specific exception types
- [ ] Implement finally block (always executed)
- [ ] Handle raise statements
- [ ] Support nested try blocks
- [ ] Connect exception edges properly

### Technical Requirements
- Exception edges from try block to handlers
- Finally block execution on all paths
- Proper flow after exception handling
- Re-raise handling

### Files to Modify
- `internal/analyzer/cfg_builder.go`

### Test Cases
- Try/except basic
- Multiple except handlers
- Try/except/finally
- Nested try blocks
- Raise in except clause

### Success Criteria
- Exception flow correctly modeled
- Finally blocks always reachable
- All exception paths covered

PR #673, opened 2026-07-23
migrate: adopt core LSH, LCOM, DFA, and semantic evidence from polyscan

DaisukeYoda:
Part of the Phase 3 adoption tracked in [polyscan#12](https://github.com/ludo-technologies/polyscan/issues/12) — the `lsh, lcom, dfa, semantic` checklist item, following the clone (#664), APTED (#669), CFG (#670), and graph (#672) adoptions.

## What moves to core

- **lsh**: `minhash.go` deleted (core's MinHash math is bit-identical: same seed, hash family, banding keys). `lsh_index.go` becomes the same thin candidate-index adapter jscan uses — int fragment IDs, deterministic sorted output, `defaultLSHMaxCandidates` cap.
- **lcom**: the LCOM4 union-find (shared-variable + call-edge unions) and risk assessment now come from `core/lcom.ComputeLCOM4` / `AssessRisk`. Python method/attribute extraction — decorator exclusion, `@property` reclassification, ctypes `_fields_` handling — stays local, as do pyscn's result semantics (excluded methods counted in `TotalMethods`, ctypes-declared fields in `InstanceVariables`, LCOM4=1 for method-less classes).
- **dfa**: `dfa.go` is now aliases over `core/dfa`; def-use chain construction and reaching-def linking (including the `DefKindPattern` same-position rule) live in core. Python def/use extraction moves onto `pythonRefExtractor`, which plugs into core via the `RefExtractor` and `ParamExtractor` extension points. Kind names follow core (`DefKindParam`, `UseKindLoad`, …).
- **semantic**: evidence penalties (disjoint string literals, missing shared strong signals, incompatible return categories) now come from `core/semantic.ApplySemanticEvidence` with `SemanticSignals`; Python signal extraction stays local.

## Deliberately kept in pyscn

`DFAFeatures`/`ExtractDFAFeatures` keep pyscn's shape (pairs-per-definition chain average) rather than adopting core's per-chain metric, and `compareCFGFeatures`/`compareDFAFeatures` and the min-cyclomatic gate stay local — adopting core's variants would shift Type-4 similarity scores, which is out of scope for this migration.

## Verification

- Full test suite passes; LSH/clone paths also pass under `-race`.
- A/B run of `analyze --json` (this branch vs main) over `testdata/python`: `lcom`, `complexity`, `dead_code`, `community_analysis`, and `suggestions` sections are byte-identical; clone aggregates identical (22 clones / 10 groups / 15 pairs — member ordering churn matches main's own run-to-run nondeterminism). The only summary delta is a 1-ULP float difference in `deps_main_sequence_deviation`, a per-binary FP compilation artifact in the untouched dependency area (each binary is self-stable across runs).

Net: −989 lines.


DaisukeYoda:
Confirmed the candidate-cap regression on both axes — thanks for the repros:

1. **Perf**: the adapter capped only after core materialized, converted, and sorted every colliding candidate, so dense buckets cost O(total candidates) per query.
2. **Selection**: sort-then-truncate keeps the N smallest IDs instead of the first candidates in band traversal order, so previously-verified pairs (e.g. the 2-band/cap-1 case) can drop out.

Fix chain:

- Core now provides a capped, order-guaranteed traversal — `FindCandidatesLimit` walks bands in order and stops at the cap before materializing: [polyscan#20](https://github.com/ludo-technologies/polyscan/pull/20). That PR also fixes jscan's adapter, which had the same regression, and adds band-order-selection tests in core. Dense-bucket benchmark (8k fragments, cap 1024): 3.37 ms / 1.43 MB per query → 75 µs / 178 KB.
- This branch has a follow-up commit ready (adapter delegates the cap to `FindCandidatesLimit`; tests now assert *which* candidates survive the cap, not just the count — e.g. IDs added [10, 20, 1] with cap 2 must select [10, 20], not [1, 10]). It will be pushed with the `go.mod` bump once polyscan#20 lands on main, since the pseudo-version needs a reachable commit.

PR #675, opened 2026-07-23
Fix module-qualified enum coupling

### LINKED ISSUE #667: [BUG][auto] cbo-enum-member-inflation-module-qualified: CBO still double-counts enum members accessed via a module alias
## Description

Issue #593 / PR #594 fixed CBO's enum-member double-counting for the *direct-import* case (`from mod import RuleMode` then `RuleMode.BLOCKING`), where the import binding itself is the enum class name. But the same double-counting bug still occurs for the **module-qualified** case: `import mod as alias` (or `from pkg import _enums as reflection`) then `alias.EnumClass.MEMBER`. Here the enum class (`alias.EnumClass`) and each enum member access (`alias.EnumClass.MEMBER`) are still counted as separate `DependentClasses` entries, because `callDependencyName`'s enum-member guard only checks whether `parts[0]` (the top-level import binding) itself looks like a class reference — it doesn't recognize that `parts[0].parts[1]` (a two-level module.Class prefix) is the actual class, with any further segment being a member access on it.

## Repro

```python
# pkg/_enums.py
import enum
class Cardinality(enum.Enum):
    One = "One"
    Many = "Many"

# main.py
from pkg import _enums as reflection

class Foo:
    def bar(self, card):
        if card == reflection.Cardinality.One:
            return 1
        elif card == reflection.Cardinality.Many:
            return 2
        return 0
```

```bash
./pyscn analyze --json --no-open --select cbo .
```

## Expected

`Foo`'s coupling to `reflection.Cardinality` should be counted once (`CouplingCount: 1`).

## Actual

```json
{"Name": "Foo", "Metrics": {"CouplingCount": 3, "DependentClasses": ["reflection.Cardinality", "reflection.Cardinality.Many", "reflection.Cardinality.One"]}}
```
Reproduced against current `main` (commit `37793ce`, pyscn `v1.26.4-9-g37793ce`) — i.e. after PR #594 merged, confirming the fix didn't cover this access pattern.

Found in the wild auditing `geldata/gel-python@8a956d6`: `GeneratedSchemaModule` in `gel/_internal/_codegen/_models/_pydantic.py` is reported with `CouplingCount: 22`, but 9 of the 22 `DependentClasses` entries are enum-member accesses (`reflection.Cardinality.{One,Many,AtMostOne}`, `reflection.PointerKind.{Link,Property}`, `reflection.CallableParamKind.{NamedOnly,Positional,Variadic}`, `reflection.TypeModifier.Optional`) double-counted alongside their already-counted parent enum classes (`reflection.Cardinality`, `reflection.PointerKind`, `reflection.CallableParamKind`, `reflection.TypeModifier`). The real distinct-type count is 13, not 22 — a 41% inflation that pushes this class into the "high coupling" risk tier.

## Root cause

`internal/analyzer/cbo.go`, `callDependencyName`: the enum-member guard
```go
if imported && len(parts) > 1 {
    if binding := parts[0]; a.isImportedDependency(binding) && a.looksLikeClassReference(binding) {
        return binding
    }
    ...
}
```
only fires when `parts[0]` itself (the direct import binding) reads as a class name. For `reflection.Cardinality.One`, `parts[0]` is `"reflection"` — a lowercase module alias, not a class — so the guard doesn't match, and the generic longest-prefix-first loop below falls through to accepting the full 3-segment `reflection.Cardinality.One` as its own "class reference" (since `dependencyLeafName` sees the CapWords-style leaf `One`).

## Priority: P2

Visible but bounded inaccuracy: coupling counts are inflated for any class that accesses enum members through a module-qualified alias (a very common import style in typed Python codebases, e.g. `import foo.enums as enums; enums.Color.RED`), pushing classes across risk-tier thresholds and triggering false "reduce coupling" suggestions — same failure mode as #593, just a different access pattern that its fix didn't cover.

## Found via

FP-audit skill in repo `geldata/gel-python@8a956d62b380db36ecaacc57952a94b1bab40d6a`, pyscn version `v1.26.4-9-g37793ce`.

ColumbusLabs:
## Summary

- collapse module-qualified enum member reads to their owning enum class for CBO accounting
- preserve existing handling for direct imports, module constants, constructors, class methods, and nested dotted constructors
- add focused regressions for the reported `reflection.Cardinality.One` case and neighboring behaviors

## Root cause

`callDependencyName` recognized enum members when the imported binding itself looked like a class, but not when a lowercase imported module alias was followed by a class and member. The generic prefix heuristic then treated the member as another class dependency.

## Verification

- focused CBO analyzer tests
- `go test ./...`
- `go test -race ./... -count=1`
- `go vet ./...`
- `golangci-lint run`
- `git diff --check`

Fixes #667

### DIFF
--- a/internal/analyzer/cbo.go
+++ b/internal/analyzer/cbo.go
@@ -883,6 +883,12 @@ func (a *CBOAnalyzer) callDependencyName(className string, allClasses map[string
 		if binding := parts[0]; a.isImportedDependency(binding) && a.looksLikeClassReference(binding) {
 			return binding
 		}
+		if suppressUppercaseLeaf && len(parts) > 2 {
+			owner := strings.Join(parts[:len(parts)-1], ".")
+			if a.looksLikeClassReference(owner) {
+				return owner
+			}
+		}
 		if suppressUppercaseLeaf && isUppercaseConstant(parts[len(parts)-1]) {
 			return ""
 		}
--- a/internal/analyzer/cbo_test.go
+++ b/internal/analyzer/cbo_test.go
@@ -1315,6 +1315,65 @@ class HardcodedRDSPasswordRule:
 	assert.Equal(t, 2, hardcoded.CouplingCount)
 }
 
+func TestCBOAnalyzer_ModuleQualifiedEnumMembersCollapseToEnumClass(t *testing.T) {
+	pythonCode := `
+from pkg import _enums as reflection
+
+class Foo:
+    def bar(self, card):
+        if card == reflection.Cardinality.One:
+            return 1
+        elif card == reflection.Cardinality.Many:
+            return 2
+        return 0
+`
+
+	ast, err := parseCode(pythonCode)
+	require.NoError(t, err)
+
+	options := DefaultCBOOptions()
+	options.GroupNamespaceImports = false
+	results, err := NewCBOAnalyzer(options).AnalyzeClasses(ast, "main.py")
+	require.NoError(t, err)
+	require.Len(t, results, 1)
+
+	foo := results[0]
+	assert.Equal(t, "Foo", foo.ClassName)
+	assert.Equal(t, []string{"reflection.Cardinality"}, foo.DependentClasses)
+	assert.Equal(t, 1, foo.CouplingCount)
+}
+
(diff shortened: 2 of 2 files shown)


DaisukeYoda (approved):
Thanks! LGTM!

PR #686, opened 2026-07-25
Add per-module quality hotspot metrics

### LINKED ISSUE #595: [ENH][auto] per-module-aggregate-metrics: no per-module quality aggregation in any output format
## Summary

pyscn provides excellent per-function complexity detail (96 function entries in this repo's analyze.json) but **no per-module rollup** of complexity, cognitive, or dead-code metrics. `system.DependencyAnalysis.ModuleMetrics` has structural fields (LinesOfCode, FunctionCount, coupling) but not quality aggregates (avg complexity, cognitive, findings-per-module). A reviewer must manually aggregate 96 per-function entries to identify hotspot modules.

## Rationale (user value)

A team lead or architect reviewing pyscn output cannot quickly identify which files/modules are the worst quality offenders. `aw/api_endpoints/job.py` (694 lines, 9 functions) is a large file but the report gives no module-level summary — the reviewer must manually scan JSON.

Per-module quality aggregation is the natural bridge between per-function detail and project-level summary. It enables:
- File-by-file comparison for code review prioritization
- Hotspot identification without manual aggregation
- Drill-down from module → function for focused remediation

## Evidence

From this repo's analyze.json (35,732 lines, 106 files):

1. **`complexity.Functions`**: 96 per-function entries with `FilePath`, `StartLine`, `Metrics.Complexity`, `Metrics.CognitiveComplexity` — but no per-file rollup
2. **`complexity.Summary`**: Project-level aggregates only (`AverageComplexity: 7.52`, `FilesAnalyzed: 106`) — not distributed to modules
3. **`system.DependencyAnalysis.ModuleMetrics`**: 106 entries with `LinesOfCode`, `FunctionCount`, `ClassCount`, coupling metrics — but NO `AvgComplexity`, `AvgCognitive`, `MaxComplexity`, `HighRiskFunctionCount` fields
4. **`dead_code.files`**: `null` (no dead code in this repo), but even if present, no per-module dead-code aggregation exists in `ModuleMetrics`

Example: `aw/api_endpoints/job.py` — 694 lines, 9 functions detected by complexity. Max complexity 14 at line 253-268. No per-module entry in the JSON surfaces this file as a hotspot.

## Root cause

Three layers of infrastructure gap:

1. **`domain/system_analysis.go:127-157`** (`ModuleDependencyMetrics` struct): Has `LinesOfCode`, `FunctionCount`, coupling fields — but no `AvgComplexity`, `AvgCognitiveComplexity`, `MaxComplexity`, `HighRiskFunctionCount` fields
2. **`service/system_analysis_service.go:1606-1667`** (`extractModuleMetrics()`): Copies coupling metrics from `internal/analyzer.ModuleMetrics` but never reads `CyclomaticComplexity` (which exists at `internal/analyzer/dependency_graph.go:111`)
3. **`app/analyze_usecase.go:451-466`** (`SystemAnalysisRequest`): `ComplexityData`/`DeadCodeData`/`ClonesData` maps are initialized as empty and never populated from parallel analysis task results — the wiring between detectors and the system analysis is incomplete

The internal analyzer infrastructure (`ModuleMetrics.CyclomaticComplexity` at `dependency_graph.go:111` and the read path at `coupling_metrics.go:106-108`) exists but is unreachable because the bridging data maps are never populated.

## Proposed change

1. Add quality-aggregation fields to `domain.ModuleDependencyMetrics`: `AvgComplexity`, `AvgCognitiveComplexity`, `MaxComplexity`, `HighRiskFunctionCount`, `DeadCodeBlockCount`, `ExceptionHandlerCount`
2. Populate `SystemAnalysisRequest.ComplexityData`/`DeadCodeData` in `app/analyze_usecase.go` after parallel analysis tasks complete, keyed by module file path
3. Extend `extractModuleMetrics()` to read and copy these fields
4. Add a `module_quality` or `file_metrics` top-level key to analyze.json with per-file rollups
5. Render a sortable table in the HTML output showing per-module quality metrics

## Impact / Effort / Confidence

- **Impact**: high — bridges the gap between per-function detail and project-level summary; enables hotspot identification
- **Effort**: medium — requires struct changes + data wiring + output formatting
- **Confidence**: high — gap confirmed at all three layers (domain/service/app)

## Verification

Survived refutation — confirmed that no per-module complexity/cognitive/dead-code aggregation exists in any output format (JSON, HTML, text). The internal `ModuleMetrics.CyclomaticComplexity` field exists but is never bridged to the domain layer. No `--by-module` flag exists. No `module_quality` top-level key in analyze.json.

## Repro

```python
# two_files/module_a.py — high complexity
def f1(a, b, c, d):
    if a > 0:
        if b > 0:
            if c > 0:
                if d > 0:
                    return 1
    return 0

# two_files/module_b.py — low complexity
def g(x):
    return x + 1
```

Run `pyscn analyze --json two_files/`. The JSON will have 2 per-function entries in `complexity.Functions` but no `module_quality` key showing module_a has avg_complexity=5 vs module_b=1.

## pyscn version

v1.24.3-14-g67b243f

---

Found via the pyscn-enhancement-audit skill in repo `O-X-L/ansible-webui@ffd298e6f85dcda1c321de1a9c9383f4d6b194c7`.

M-Hassan-Raza:
Closes #595

This adds stable per-module quality rollups across complexity, cognitive complexity, exception handling, and dead-code analysis.

The rollups are calculated before presentation filters, so changing min_complexity, report_unchanged, or dead-code severity only affects displayed findings, not the module aggregates. Module identity is normalized only at the app-level join, while analyzer output keeps the caller-facing path.

The unified analyze output now includes module_quality in JSON and YAML, adds the top hotspots to text and CSV output, and includes a sortable Modules table in the HTML report.

Review converged on module_quality as the single public owner for these aggregates. The existing SystemAnalysisRequest and dependency metric contracts remain unchanged, avoiding duplicate payload fields and competing writers.

Compatibility notes:

- Existing Go APIs and analyzer path output are unchanged.
- JSON and YAML gain the additive module_quality collection.
- analyze --csv includes module-quality rows after the existing summary, so consumers assuming a fixed row count need to account for the new rows.

Validation:

go test ./...
go vet ./...
go test -race ./...
golangci-lint run

I also compared JSON output across different complexity thresholds, verified equivalent input paths only produce one module, and checked both mixed and system-only analysis preserve relative paths. The sortable HTML table was checked in a real browser.

Back after a bit of a hiatus — feels good to pick up a properly chunky issue again 😁


app/analyze_usecase.go
```diff
@@ -283,6 +284,10 @@ func (uc *AnalyzeUseCase) execute(ctx context.Context, useCaseCfg AnalyzeUseCase
 	if len(files) == 0 {
 		return nil, fmt.Errorf("no Python files found in the specified paths")
 	}
+	files, err = canonicalAnalysisPaths(files)
+	if err != nil {
+		return nil, err
+	}
```
DaisukeYoda: This canonicalization is what the module join needs, but it also changes the paths every analyzer reports: `analyze` output now carries absolute, machine-specific paths instead of the paths the user passed. I confirmed this against main. It makes reports non-portable and non-diffable, exposes local directory layout, and breaks downstream consumers for a reason unrelated to module rollups. Normalizing only at the join boundary would keep this contained.

service/system_analysis_service.go
```diff
@@ -376,7 +376,9 @@ func (s *SystemAnalysisServiceImpl) buildDependencyAnalysisResult(ctx context.Co
 	}
 
 	// Calculate coupling metrics
-	metricsCalculator := analyzer.NewCouplingMetricsCalculator(graph, analyzer.DefaultCouplingMetricsOptions())
+	metricsOptions := analyzer.DefaultCouplingMetricsOptions()
+	metricsOptions.ComplexityData = req.ComplexityData
```
DaisukeYoda: These request-level complexity and dead-code inputs are never populated anywhere outside tests, so both this and the dead-code loop below can only ever iterate empty maps. The real values arrive later via the system-publishing step, which overwrites the same fields wholesale. That's two writers for one set of fields with one of them dead

domain/system_analysis.go
```diff
@@ -146,9 +145,11 @@ type ModuleDependencyMetrics struct {
 	Distance         float64 // D - distance from main sequence
 
 	// Quality metrics
-	Maintainability float64   // Maintainability index (0-100)
-	TechnicalDebt   float64   // Estimated technical debt in hours
-	RiskLevel       RiskLevel // Overall risk assessment
+	Maintainability         float64   // Maintainability index (0-100)
+	TechnicalDebt           float64   // Estimated technical debt in hours
+	RiskLevel               RiskLevel // Overall risk assessment
+	ModuleComplexityMetrics `yaml:",inline"`
```
DaisukeYoda: Embedding here means the same metrics ship twice in one payload, once in the module collection and once per dependency metric. For the dead-code count it's actually a third copy, since per-file `total_findings` already exists. One home would be easier to keep consistent.

domain/module_quality.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		modules[key] = module
+	}
+	return modules
+}
+
+// AggregateModuleQuality combines analyzer-owned, pre-filter rollups with
+// structural module metadata. Unified analysis gives every task the same
+// canonical absolute paths, so joins use exact cleaned identities.
+func AggregateModuleQuality(response *AnalyzeResponse) []ModuleQualityMetrics {
```
DaisukeYoda: This reads the whole analyze response, and the function below mutates another analyzer's response. That's use-case orchestration living in the entity package. Moving the assembly to the use case would keep domain free of cross-analysis wiring.

service/complexity_service.go
```diff
@@ -69,6 +69,7 @@ func (s *ComplexityServiceImpl) Analyze(ctx context.Context, req domain.Complexi
 
 	// Filter and sort results
 	functionsParsed := len(allFunctions)
+	moduleRollups := domain.AggregateComplexityByModule(allFunctions)
```
DaisukeYoda, line 71: Now that the report check moved out of calculation, this count includes complexity-1 functions when `report_unchanged` is off, which changes the user-visible reported-versus-parsed figure. Intentional or not, it belongs in the behavior-change list.

DaisukeYoda (changes requested):
@M-Hassan-Raza 

Thank you for coming back! This is a genuinely useful addition from a user's perspective.

I left a few comments on things I noticed. One of them, the path change, I'd like resolved before this goes in. Please take a look!

app/analyze_usecase.go
M-Hassan-Raza: Resolved in 77af996 and 9f60bd7. Collected paths now retain the first caller-facing spelling; an app-owned path index uses absolute cleaned identities only for deduplication and joining. Added mixed-analysis and system-only relative-path regressions, plus a real CLI check.

service/system_analysis_service.go
M-Hassan-Raza: Agreed. 0043feb removes the request-level quality wiring and the later system publisher, and restores the existing SystemAnalysisRequest/coupling contracts unchanged. The unified response now has one post-parallel assembly path.

domain/system_analysis.go
M-Hassan-Raza: Agreed. 0043feb removes the added quality fields from ModuleDependencyMetrics. Top-level module_quality is now the sole new owner of the aggregate contract.

domain/module_quality.go
M-Hassan-Raza: Moved in 0043feb. Cross-analysis joining and ordering now live in app/module_quality.go; domain keeps only the typed metric contracts and analyzer-owned reducers. There is no longer any cross-response mutation.

service/complexity_service.go
M-Hassan-Raza: Resolved in e9f9e6f. The reporting filter now returns the visible population and the legacy pre-minimum count after report_unchanged, preserving FunctionsParsed behavior. Module rollups still consume the complete pre-presentation-filter population, with regression coverage for both contracts.

DaisukeYoda (approved):
Thanks! LGTM!

PR #695, opened 2026-07-29
fix: make dependency analysis scale

### LINKED ISSUE #689: [BUG] deps analysis is exponential: 44 three-line files hang >90s, Django source never completes
## Summary

`calculateDepthFromModule` in the dependency (`deps`) analysis enumerates **all simple paths** in the module dependency graph instead of memoizing per-node depth. Runtime is exponential in the graph's branching structure, not in file count.

A synthetic package of **44 Python files, 3 lines each**, hangs for over 90 seconds. Real-world impact: `pyscn analyze` on the Django source tree (`django/django`, 908 files) ran for **over 30 minutes** without completing, pinned at 99% CPU on a single core.

This affects the **default** `pyscn analyze` invocation, since `deps` runs unless `--skip-deps` is passed.

## Minimal reproduction

Each module imports the next two, producing Fibonacci-many simple paths:

```bash
gen() {
  rm -rf pkg$1; mkdir -p pkg$1; touch pkg$1/__init__.py
  n=$1; i=0
  while [ $i -lt $n ]; do
    { [ $((i+1)) -lt $n ] && echo "from . import m$((i+1))"
      [ $((i+2)) -lt $n ] && echo "from . import m$((i+2))"
      echo "def f$i(): return $i"; } > pkg$1/m$i.py
    i=$((i+1))
  done
}

for n in 32 36 40 44; do gen $n; time pyscn analyze pkg$n --select deps --json --no-open; done
```

Measured (Apple Silicon, v1.29.0 at 81b7ec0):

| modules / .py files | wall time |
|---|---|
| 32 | 1.08s |
| 36 | 7.1s |
| 40 | 49.9s |
| 44 | >90s (killed) |

Growth is about 7x per 4 added modules, matching φ⁴ ≈ 6.85 for the Fibonacci path count. Every file is three lines long.

## File count is not the driver

Sub-packages of Django, `--select deps` only:

| target | .py files | wall time |
|---|---|---|
| `django/http` | 5 | 31ms |
| `django/utils` | 48 | 89ms |
| `django/core` | 112 | 178ms |
| `django/db` | 123 | 880ms |
| `django/contrib` | 338 | 416ms |
| `django/django` (whole) | 908 | **>30min, did not finish** |

`contrib` has almost 3x the files of `db` but runs in half the time. What matters is how many distinct paths the import graph admits.

## Root cause

`service/system_analysis_service.go:1448-1469`:

```go
func (s *SystemAnalysisServiceImpl) calculateDepthFromModule(graph *analyzer.DependencyGraph, current string, visited map[string]bool, currentDepth int) int {
	if visited[current] {
		return currentDepth
	}
	visited[current] = true
	defer delete(visited, current)   // <-- stack-scoped only, no per-node memo
	...
	for dep := range node.Dependencies {
		depth := s.calculateDepthFromModule(graph, dep, visited, currentDepth+1)
```

`visited` is scoped to the current recursion stack and cleared on unwind, which correctly prevents infinite recursion through cycles but provides no memoization. Any module reachable by more than one path (a shared `utils`, `common`, or `__init__`) has its entire subgraph re-explored once per incoming path. This is the classic longest-path-without-DP mistake.

`calculateMaxDepth` (`:1435-1446`) then launches this from **every** node, multiplying the cost by |V|.

### Confirmed by profiling

Sampling the hung 908-file run (built without `-s -w` so symbols resolve):

```
289  service.(*SystemAnalysisServiceImpl).calculateDepthFromModule
 69  service.(*SystemAnalysisServiceImpl).calculateDepthFromModule.deferwrap1
  4  service.(*SystemAnalysisServiceImpl).calculateMaxDepth
  4  service.(*SystemAnalysisServiceImpl).buildDependencyAnalysisResult
```

289 frames of self-recursion. The process reached `calculateMaxDepth` (the last phase of `buildDependencyAnalysisResult`, line 417) within the first minute, so every preceding phase completed quickly and this single function accounts for the entire remaining runtime. RSS stayed at ~10MB, consistent with pure integer recursion.

## Two related problems in the same area

**1. The `maxPaths` cap in `findPathsFromModule` is dead code** (`service/system_analysis_service.go:1392-1397`):

```go
func (s *SystemAnalysisServiceImpl) findPathsFromModule(..., maxPaths int) []domain.DependencyPath {
	var paths []domain.DependencyPath

	if len(paths) >= maxPaths {   // paths was just declared; len is always 0
		return paths
	}
```

The guard tests a slice that is always empty at function entry, so it never fires. The check at line 1426 only runs after a full recursive branch has already unwound. The intended "top 10 chains" bound does not actually bound the search.

**2. `CouplingMetricsCalculator.findMaxDepthFrom` has the same unmemoized shape** (`internal/analyzer/coupling_metrics.go:319-341`), using `defer func() { visited[module] = false }()`. It is only launched from root modules, so it is less severe, but it is the same latent bug.

## The 10-minute timeout does not help

`cmd/pyscn/analyze.go:191` wraps the command in a 10-minute `context.WithTimeout`, but `buildDependencyAnalysisResult` only polls `ctx.Err()` *between* phases (lines 356, 374, 384, 390, 396, 402). None of the recursive functions check the context internally, so once one starts blowing up the timeout cannot interrupt it. The observed 30+ minute run is direct evidence.

## Suggested fix

1. Replace `calculateDepthFromModule` with a memoized longest-path computation. The dependency graph already has SCC detection available (`circularDetector.DetectCircularDependencies()` runs earlier in the same function and uses Tarjan). Condense to the SCC DAG, then a single-pass DP gives max depth in O(V+E) instead of O(paths).
2. Apply the same treatment to `findMaxDepthFrom` in `coupling_metrics.go`.
3. Fix the dead guard in `findPathsFromModule`, and bound the enumeration properly (cap total paths explored, not just paths returned).
4. Add `ctx` checks inside the recursive functions so the existing timeout can actually take effect.
5. Consider a regression test using the generator above at N=44, asserting the analysis completes in well under a second.

## Environment

- pyscn v1.29.0 (commit 81b7ec0)
- go1.24.6, darwin/arm64

M-Hassan-Raza:
Fixes #689.

Replaces exhaustive path enumeration with a cancellable SCC-condensed topology pass shared by depth, coupling, and bounded longest-chain analysis. Lazy-only imports remain available to coupling and architecture checks but are excluded from load-time topology. Django dependency analysis completes in about 1.2s across 908 Python files.

Breaking behavior: cycles now count as one component, so MaxDepth, dependency health scores, and LongestChains can change for cyclic graphs.

Validated with tests, race detector, vet, lint, and the strict docs build.

### DIFF
--- a/docs/algorithms/dependency.md
+++ b/docs/algorithms/dependency.md
@@ -202,21 +202,26 @@ Dependency depth measures the length of the longest chain of transitive dependen
 
 ### Calculation
 
-The algorithm performs a recursive DFS from each root module (modules with `OutDegree == 0`), tracking the maximum depth reached. Cycles are handled by maintaining a visited set to prevent infinite recursion.
+The analyzer first condenses every strongly connected component (SCC) into one
+node. It then calculates the longest path through the resulting directed
+acyclic graph in a single topological pass.
 
 ```
-function findMaxDepthFrom(module, visited):
-    if module in visited: return 0
-    visited.add(module)
+components = stronglyConnectedComponents(dependencyGraph)
+dag = condense(dependencyGraph, components)
+depth = zeroes(len(components))
 
-    maxChildDepth = 0
-    for each dependency of module:
-        childDepth = findMaxDepthFrom(dependency, visited)
-        maxChildDepth = max(maxChildDepth, childDepth)
+for component in topologicalOrder(dag):
+    for dependency in dag.dependencies(component):
+        depth[dependency] = max(depth[dependency], depth[component] + 1)
 
-    visited.remove(module)
-    return maxChildDepth + 1
+maxDepth = max(depth)
 ```
+
+Depth counts edges between components. A project with one module has depth
+zero; `A -> B -> C` has depth two. Modules in a cycle form one component
+because the circular dependency is reported separately. The calculation is
+`O(V + E)` in the number of modules and dependencies.
 
 ### Expected Depth
 
@@ -329,7 +334,7 @@ The `SystemMetrics` struct aggregates module-level metrics into project-wide ind
 | Average Abstractness | sum(A) / N | System-wide average abstractness |
 | Main Sequence Deviation | sum(D) / N | Average distance from main sequence |
 | Cyclic Dependencies | count of modules in SCCs | Number of modules involved in cycles |
-| Max Dependency Depth | longest chain length | Maximum transitive dependency depth |
+| Max Dependency Depth | longest SCC-DAG path in edges | Maximum transitive dependency depth |
 
 ### Modularity Index
 
--- a/docs/algorithms/dependency.md
+++ b/docs/algorithms/dependency.md
@@ -228,7 +228,8 @@ keeps the top paths from each component, so the reported top 10 are ranked
 globally rather than selected by traversal order. Component transitions are
 expanded back into real module-to-module dependency paths. When a path crosses
 a cyclic component, the analyzer uses a deterministic internal path between
-the incoming and outgoing modules.
+the incoming and outgoing modules. Coupling metrics consume the same topology
+result instead of rebuilding or traversing the dependency graph again.
 
 ### Expected Depth
 
--- a/docs/algorithms/dependency.md
+++ b/docs/algorithms/dependency.md
@@ -223,6 +223,13 @@ zero; `A -> B -> C` has depth two. Modules in a cycle form one component
 because the circular dependency is reported separately. The calculation is
 `O(V + E)` in the number of modules and dependencies.
 
+Longest-chain reporting uses the same condensed DAG. A bounded dynamic program
+keeps the top paths from each component, so the reported top 10 are ranked
+globally rather than selected by traversal order. Component transitions are
+expanded back into real module-to-module dependency paths. When a path crosses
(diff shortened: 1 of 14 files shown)


DaisukeYoda:
Nice fix, the whole Django tree finishes in about 1.2s locally. One thing before merge: the condensation counts lazy function-body imports while cycle reporting excludes them, so cycle-free repos still get their depth collapsed and their score inflated (django/contrib: no cycles either way, depth 17 to 11, score 70 to 80). Matching the cycle detector's edge filter would fix it.

M-Hassan-Raza:
Good catch, fixed in 29115cb. Topology now uses the same load-time edge contract as cycle detection, while lazy edges remain in coupling, dependency matrices, and architecture checks. Lazy-to-eager promotion also invalidates stale topology. django/contrib is back to dependency score 70, and the full Django tree still finishes in about 1.2s. Full tests, race, vet, lint, and docs are green.

DaisukeYoda (approved):
Thanks!

M-Hassan-Raza:
@DaisukeYoda Thanks for merging timely! I have a TON of projects in Django so the next release will help me optimize my codebases as well :D

PR #697, opened 2026-07-31
fix: --min-complexity filters the analysed function set, not just the report:...

### LINKED ISSUE #696: [BUG] --min-complexity filters the analysed function set, not just the report: raising it lowers the health score
## Summary

`--min-complexity` is documented as `Minimum complexity to report`, which reads as a display threshold. It actually filters the function list that everything downstream is computed from, including `summary.average_complexity`, `summary.complexity_score` and `summary.health_score`.

Because the average is then taken over only the functions that survived the filter, it is mathematically pinned at or above the threshold. The consequence is backwards: **raising `--min-complexity` makes your score worse, and lowering it makes your score better, on identical code.**

## Repro

One file, ten functions: one with complexity 6, nine trivial ones.

`pkg/sample.py`:
```python
def busy(n):
    if n == 0:
        return 0
    if n == 1:
        return 1
    if n == 2:
        return 2
    if n == 3:
        return 3
    if n == 4:
        return 4
    return -1

def tidy0(x):
    return x + 0

# ... tidy1 through tidy8, same shape
```

```console
$ pyscn analyze --json --no-open --select complexity pkg/sample.py
# summary.average_complexity  = 6
# summary.complexity_score    = 70
# summary.health_score        = 94
# complexity.Functions        = 1 entry

$ pyscn analyze --json --no-open --select complexity --min-complexity 1 pkg/sample.py
# summary.average_complexity  = 1.45
# summary.complexity_score    = 100
# summary.health_score        = 100
# complexity.Functions        = 11 entries
```

Same file, same code. Only the flag differs.

Real-world scale, `django/db/models/sql/query.py` at `84322a29ce`:

| | functions reported | total complexity | average_complexity | complexity_score |
| --- | ---: | ---: | ---: | ---: |
| default (`--min-complexity 5`) | 29 | 277 | 9.55 | 70 |
| `--min-complexity 1` | 105 | 425 | 4.05 | 100 |

72% of the file's functions are absent from the default report, and the score computed from what remains is 30 points worse.

## Expected

Either of these would be coherent, but the current behaviour is neither:

1. `--min-complexity` is a display threshold. It controls which functions appear in the human-readable output; `summary.*` is computed over every function in the analysed files.
2. `--min-complexity` is an analysis filter. Then the flag name and help text should say so, and the docs should note that `average_complexity` is conditional on it.

(1) seems right given the flag name and that `--min-severity` sits next to it as a genuine reporting filter.

## Second-order impact: before/after comparisons are non-monotonic

This is what led me here. If you diff two pyscn reports to see whether a change made things worse, the threshold makes the comparison lie in both directions:

- A function simplified from complexity **6 → 4** drops out of the report entirely and reads as a **deleted function**.
- A function that grows from **3 → 8** appears from nowhere, with no "before" value to compare against, so the regression goes uncounted.

Any CI gate of the form "fail if complexity increased since main" is affected, as is anything comparing two reports. We hit this in a benchmark that scores SWE-bench patches for maintainability ([polyscan/benchmarks/patch-quality](https://github.com/ludo-technologies/polyscan/tree/main/benchmarks/patch-quality)); it silently hid roughly three quarters of the complexity increases we were trying to measure until we pinned `--min-complexity 1`.

## Related

- #690, `pyscn check` exits 0 on unparseable files. Same family: something is dropped from the analysis and the score is computed as though it never existed.

yhay81:
Fixes #696.

## Summary

[BUG] --min-complexity filters the analysed function set, not just the report: raising it lowers the health score

## Validation

- Mechanical gate: +20/-10, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.


service/complexity_service_test.go
```diff
@@ -559,10 +559,12 @@ func TestComplexityService_GenerateSummary(t *testing.T) {
 
 		req := domain.ComplexityRequest{MinComplexity: 5}
 		filtered, functionsParsed := service.filterFunctions(allFunctions, req)
-		summary := service.generateSummary(filtered, 1, req, functionsParsed)
+		summary := service.generateSummary(allFunctions, len(filtered), 1, req, functionsParsed)
 
 		assert.Equal(t, 1, summary.TotalFunctions, "post-filter count is 1")
 		assert.Equal(t, 3, summary.FunctionsParsed, "pre-filter total is 3 (2 dropped by min_complexity=5)")
+		assert.InDelta(t, 11.0/3.0, summary.AverageComplexity, 1e-9, "average is over all parsed functions, not the filtered set")
```
DaisukeYoda, line 634: This asserts the helper's arithmetic given already-correct arguments, not the wiring that was actually broken. I reverted both call sites to the old behaviour and the full suite still passed. Driving the analysis entrypoint with two min-complexity values and asserting an identical summary would be the real guard, and would cover the snapshot path too.

service/complexity_service.go
```diff
@@ -72,8 +72,9 @@ func (s *ComplexityServiceImpl) Analyze(ctx context.Context, req domain.Complexi
 	filteredFunctions, functionsParsed := s.filterFunctions(allFunctions, req)
 	sortedFunctions := s.sortFunctions(filteredFunctions, req.SortBy)
 
-	// Generate summary
-	summary := s.generateSummary(sortedFunctions, filesProcessed, req, functionsParsed)
+	// Generate summary over the complete population so min_complexity only
+	// affects which functions are displayed, not the aggregate metrics.
+	summary := s.generateSummary(allFunctions, len(sortedFunctions), filesProcessed, req, functionsParsed)
```
DaisukeYoda: The fix stops one layer short. Per-directory aggregation still computes from the filtered set, so it keeps exactly this bug. A single-directory project now reports a project average and a directory average that contradict each other, side by side in the same report.

service/complexity_service.go
```diff
@@ -358,10 +359,15 @@ func (s *ComplexityServiceImpl) sortByRisk(functions []domain.FunctionComplexity
 }
 
 // generateSummary creates summary statistics.
+// functions must be the complete analyzer population: min_complexity and
+// report_unchanged are presentation filters, so averages, distribution, and
+// risk counts stay stable regardless of what is displayed.
+// reportedFunctions is the post-filter count of functions included in the output.
 // functionsParsed is the pre-filter function count (all functions parsed before min_complexity filtering).
-func (s *ComplexityServiceImpl) generateSummary(functions []domain.FunctionComplexity, filesAnalyzed int, req domain.ComplexityRequest, functionsParsed int) domain.ComplexitySummary {
+func (s *ComplexityServiceImpl) generateSummary(functions []domain.FunctionComplexity, reportedFunctions int, filesAnalyzed int, req domain.ComplexityRequest, functionsParsed int) domain.ComplexitySummary {
```
DaisukeYoda: Five positional params now, including three adjacent ints the compiler can't tell apart, and one that's no longer used at all.

DaisukeYoda (changes requested):
Verified the core fix works: averages and health score are now stable across min-complexity values. Two blocking points inline.

service/complexity_service_test.go
yhay81: Addressed in feed646.

service/complexity_service.go
yhay81: Addressed in 0a64c06.

DaisukeYoda:
@yhay81 Ready for review?

yhay81:
Yes — this is ready for review now. I also addressed the remaining per-directory aggregation issue in a17e859: directory metrics now use the complete analyzed population, matching the project summary across both the regular and snapshot paths. Regression coverage verifies that presentation filters still limit the function list without changing directory metrics.

Validation: `go test ./...` and `go vet ./app ./service ./domain`.

DaisukeYoda (changes requested):
The core fix is verified: averages and the health score now hold steady across min-complexity values on both the direct and snapshot paths, and the tests drive the real entrypoint.

Two things before I approve. The summary's function counts no longer describe the population its own averages are taken over, so with report_unchanged off the top-line count and the directory table contradict each other in the same report. And the defensive fallback in the directory attachment silently restores the exact bug this PR fixes; removing it broke only a mock, so it is accommodating a test double rather than a real caller.

DaisukeYoda:
@yhay81 Could you check the additional comment? There remains two stuffs before merging. Thanks!

yhay81:
@DaisukeYoda Thanks for the follow-up — I’ve addressed both remaining points in d4592ec (with the test expectation corrected in c4ff6d0):

- The summary counts now describe the same complete analyzed population as the averages, risk distribution, health score, and per-directory rollups. `min_complexity` and `report_unchanged` only limit the displayed `Functions` list.
- I removed the fallback from `AnalyzedFunctions` to the filtered `Functions` field. The mock now follows the production response contract, and a regression test prevents the fallback from being reintroduced.

I also merged the latest `main` and resolved the conflicts. CI is green for Go 1.22 and 1.23, including `go vet`, the full test suite, `golangci-lint`, build, and the binary smoke test.

This is ready for another review. Thanks!

DaisukeYoda (approved):

DaisukeYoda:
Thanks!

PR #701, opened 2026-08-02
fix(config): stop config discovery at the target repository root

### LINKED ISSUE #666: [BUG][auto] arch-config-ancestor-leakage-unbounded-ascent: architecture config search crosses repo boundaries with no warning
## Description

`FindConfigFileFromPath` in `internal/config/toml_loader.go` walks upward from the analyzed target path looking for `.pyscn.toml` (or `pyproject.toml` with `[tool.pyscn]`), with **no boundary check at the target's own repository root**. It only stops at the filesystem root (`parent == current`). If the analyzed directory happens to be nested inside a parent directory tree that contains an unrelated `.pyscn.toml` — e.g. a checked-out target repo living under `<pyscn-checkout>/.pyscn/audit/repos/<slug>/`, or any monorepo/nested-checkout layout — pyscn silently applies that ancestor's config to the target, with no warning that the resolved config lives outside the analyzed tree.

## Repro

This was found auditing `geldata/gel-python` (commit `8a956d6`) cloned to `.pyscn/audit/repos/geldata-gel-python` inside this pyscn checkout, which has its own dogfooding `.pyscn.toml` with:
```toml
[architecture]
enabled = true
validate_layers = true
strict_mode = true

[[architecture.layers]]
name = "presentation"
packages = ["router", "routers", ..., "controller", ...]
# ... application / domain / infrastructure layers, all generic web-app package names
```

None of gel-python's packages (`gel`, `edgedb`, `gel.cli`, `gel._internal.*`) match any of these generic layer package-name patterns, so every module falls into `"unknown"` layer. With the leaked `strict_mode = true`, every cross-module edge becomes a `strict_mode` violation — including ordinary same-package imports like `gel.__main__ -> gel.cli` and the project's own auto-generated backwards-compat shim (`edgedb/*.py`, which literally re-exports `gel/*.py`).

Result: 334 of 411 (81%) reported architecture violations, and an Architecture score of 28/100, that have nothing to do with gel-python's actual structure — they are an artifact of the pyscn repo's own unrelated config leaking across a repository boundary.

Verified independently:
- Built the current binary and ran it against two synthetic zero-config repos (no ancestor `.pyscn.toml` anywhere in the path) — both scored Architecture 100/100 with zero `strict_mode` violations, confirming `autoDetectArchitecture` correctly hardcodes `StrictMode: false` when no config is found.
- Read `internal/config/toml_loader.go:428-465` (`FindConfigFileFromPath`): the loop `for { ... parent := filepath.Dir(current); if parent == current { break }; current = parent }` has no check against the target path's own VCS root or a `.git` boundary — it will ascend arbitrarily far.
- Confirmed `edgedb/__init__.py` and `edgedb/abstract.py` in gel-python are auto-generated re-export shims (`# Auto-generated shim`), so the "violations" attributed to them are especially misleading.

## Expected

Config discovery should stop ascending at the analyzed target's own repository boundary (e.g. its nearest `.git` directory, or the originally-specified target path itself), and/or emit a visible warning when the config file resolved is outside the analyzed directory tree, so users aren't silently scored against an unrelated project's architectural conventions.

## Actual

Config search walks all the way up the filesystem tree with no boundary check, silently picking up and applying an ancestor project's `.pyscn.toml` — including `strict_mode` and layer definitions that have no relationship to the analyzed code — whenever the target happens to be nested under such a directory (monorepos, nested checkouts, or tooling that clones targets under its own working tree, as this project's own audit-automation skills currently do).

## Priority: P1

Substantially misleading output a user would act on: an entire top-level score dimension (Architecture, 28/100) and the majority of reported violations (81%) do not reflect any real property of the analyzed code, with no indication to the user that a foreign config file was used. This can silently and drastically understate code quality for any nested-checkout or monorepo-subdirectory analysis.

## Found via

FP-audit skill in repo `geldata/gel-python@8a956d62b380db36ecaacc57952a94b1bab40d6a`, pyscn version `v1.26.4-9-g37793ce`. Note: this also means the pyscn-fp-audit/pyscn-enhancement-audit skills' own practice of cloning target repos under `.pyscn/audit/repos/` inside this checkout has likely contaminated the Architecture/Dependency scores reported in *every* prior audit run this way, not just this one.

DaisukeYoda:
Fixes #666

## Problem

`FindConfigFileFromPath` (and the `findPyscnToml` / `findPyprojectToml` walkers) ascended from the analyzed path all the way to the filesystem root. A repository checked out inside another working tree — a nested clone, or a package in a monorepo — silently inherited the ancestor project's `.pyscn.toml`, including `strict_mode` and layer definitions with no relationship to the analyzed code.

In the reported case that produced an Architecture score of 28/100 and 334 of 411 violations that were artifacts of the outer repo's dogfooding config.

## Change

The upward walk now stops at the repository root containing the target — a directory holding a `.git` entry (directory for a clone, file for a worktree/submodule) — inclusive, so a repo-root config still applies when analyzing a subdirectory of that same repo. The three walkers now share one `findUpward` helper instead of three copies of the loop, and the config filenames are constants.

A target with no `.git` anywhere above it has no boundary to stop at and still searches to the filesystem root; that limitation is documented.

## Test evidence

New tests in `internal/config/toml_loader_test.go`: nested checkout ignores the ancestor config (`.git` dir and `.git` worktree file), nested checkout uses its own config, `pyproject.toml` fallback is bounded the same way, and a subdirectory of a repo still finds the repo-root config.

`go test ./...` passes. `go vet ./...` passes; `golangci-lint` cannot run in this environment (it is built against go1.24, repo targets 1.25.5) — pre-existing and unrelated.

End-to-end on a synthetic nested checkout (outer dir with a strict layer config, inner git repo with two plain modules):

| target | Architecture |
| --- | --- |
| inner repo with `.git` (after fix) | 100/100, 0 violations |
| same tree with `.git` removed (pre-fix behavior) | 83/100, ancestor `strict_mode` applied |


DaisukeYoda:
Follow-up: the boundary alone did not fix `pyscn check <target>` — thanks for catching it. Pushed f818f22.

Two target-blind discovery origins were involved:

1. `resolveCheckConfig` passed `""` as the target, so discovery started at the working directory and found the outer `.pyscn.toml` before the walk could ever reach the target's own `.git`. It now passes `getTargetPathFromArgs(args)`.

2. Fixing (1) alone was still not enough. With no config inside the target, `resolveCheckConfig` returns `""`, each sub-check passes an empty `ConfigPath`, and the use cases fall back to `LoadDefaultConfig()` — which re-discovered from the working directory and leaked the outer config anyway. `LoadDefaultConfig` and `GetDefaultCloneConfig` now take the analyzed path (derived from the request's own `Paths`), and the eight duplicated `FindDefaultConfigFile` bodies delegate to one `discoverConfigFile` helper.

This also improves MCP: the server's `ConfigPath` is empty unless `PYSCN_CONFIG` is set, so discovery previously ran from the server process's working directory, which is generally unrelated to the analyzed `path`.

Verified on a synthetic nested checkout (outer project gating complexity at 3, inner git repo with a complexity-5 function), run from the outer directory:

| command | before | after |
| --- | --- | --- |
| `check --select complexity ./inner` | fails: `5 > 3` (outer config) | passes (built-in defaults) |
| `check --select complexity .` (outer's own tree) | fails: `5 > 3` | unchanged |
| `check --select complexity ./app` (subdir of outer repo) | fails: `5 > 3` | unchanged |
| `check --select complexity ./inner` (inner has own config) | — | inner's config applies |

New regression tests: `TestCheckIgnoresAncestorConfigAcrossRepositoryBoundary` and `TestCheckUsesRepositoryRootConfigForSubdirectory` in `cmd/pyscn/new_commands_test.go` (both confirmed failing against the previous commit), plus `TestConfigurationLoader_LoadDefaultConfigDiscoversFromTarget` in `service/config_loader_test.go`. `go test ./...` and `go vet ./...` pass.

Issue #677, opened 2026-07-24
[BUG][auto] syntax-error-file-analyzed-silently: Python 2 `except A, B:` bypasses the parse-error check and scores as clean


DaisukeYoda:
## Summary

pyscn's syntax-error detection does not fire for Python 2 style `except A, B:`. The file is analyzed as if valid, functions are extracted from the recovered parse, and the report gives it a clean bill of health with no error and no warning.

This is not a general "pyscn ignores broken files" problem — the parse-error path works. On a file of pure garbage, `complexity.Errors` correctly contains `Parse error: syntax errors found in source code`. It is specifically the Python 2 exception-list form that slips through, most likely because the tree-sitter Python grammar still accepts it and so no `ERROR` node is produced for the syntax check to find.

## Repro

`broken.py` (5 lines, rejected by CPython):

```python
def f(x):
    try:
        return x
    except OSError, TypeError:
        return None
```

```console
$ python3 -c "import ast; ast.parse(open('broken.py').read())"
SyntaxError: multiple exception types must be parenthesized

$ pyscn analyze --json --no-open .
Health Score: 97/100 (Grade: A)
  Complexity:     100/100 ✅  (avg: 1.5, high-risk: 0 functions)
  Dead Code:      100/100 ✅  (0 issues, 0 critical)
```

## Expected

The file is not valid Python 3. pyscn should either report a parse error for it (as it already does for other malformed input) or skip it and count it in `skipped_files` — either way the user should be able to tell that the file was not soundly analyzed.

## Actual output

From the generated `analyze_*.json` for the repro above:

```json
{
  "total_files": 1,
  "analyzed_files": 1,
  "skipped_files": 0,
  "total_functions": 2,
  "functions_parsed": 2
}
```

`complexity.Errors`, `complexity.Warnings`, `dead_code.errors` and `dead_code.warnings` are all `null`, and `complexity.Functions` contains two recovered entries:

```json
[
  {"Name": "f",        "StartLine": 1, "EndLine": 5, "Metrics": {"Complexity": 2, "ExceptionHandlers": 1}},
  {"Name": "<module>", "StartLine": 1, "EndLine": 6, "Metrics": {"Complexity": 1}}
]
```

Contrast with a file of pure garbage (`this is (((( not python at all ]]]] def :::`), where the same field is populated as it should be:

```json
"Errors": ["[.../garbage.py] Parse error: syntax errors found in source code"]
```

## Impact in the wild

[...]

PR #703, opened 2026-08-06
fix(check): fail the gate on files that cannot be parsed

### LINKED ISSUE #690: [BUG] pyscn check passes and exits 0 on code with syntax errors; unparseable files score as Grade A
## Summary

When a file fails to parse, pyscn detects it and prints a warning, but the file is then dropped from the analysis entirely and **nothing downstream reflects it**: the health score, the letter grade, the per-category subscores, and the process exit code are all computed as if the file did not exist.

The practical consequence is that `pyscn check` reports **"✅ Code quality check passed" and exits 0** on a codebase containing Python that CPython refuses to parse. Used as a CI quality gate, it passes.

This is **not** a duplicate of #677. That issue is about detection failing for one specific Python 2 form (`except A, B:`) so no error is raised at all. Here detection works correctly — `complexity.Errors` gets the right entry and a warning is printed — and the score and gate still ignore it.

## Repro

```
pkg/good.py
pkg/broken.py
```

`pkg/good.py`:
```python
def tidy(n):
    if n > 0:
        return n
    return -n
```

`pkg/broken.py` (rejected by CPython: `SyntaxError: invalid syntax`):
```python
def loader(objs):
    total = 0
    for o in objs:
        total += o
    return 0 objs
```

```console
$ python3 -c "import ast; ast.parse(open('pkg/broken.py').read())"
SyntaxError: invalid syntax

$ pyscn check pkg
🔍 Running quality check (complexity, deadcode, clones)...
Warning: Failed to parse file pkg/broken.py: syntax errors found in source code
✅ Code quality check passed
$ echo $?
0

$ pyscn analyze pkg --no-open
Warning: Failed to parse file pkg/broken.py: syntax errors found in source code

📊 Analysis Summary:
Health Score: 90/100 (Grade: A)

📈 Detailed Scores:
  Complexity:     100/100 ✅  (avg: 1.5, high-risk: 0 functions)
  Dead Code:      100/100 ✅  (0 issues, 0 critical)
  Duplication:    100/100 ✅  (0.0% fragments cloned, 0 groups)
  Coupling (CBO): 100/100 ✅  (avg: 0.0, 0/0 high-coupling)
  Cohesion (LCOM):100/100 ✅  (avg: 0.0, 0/0 high-lcom)
  Architecture:   100/100 ✅  (100% compliant)

[Badge] Add this to your README to show off your score:
[![pyscn quality](https://img.shields.io/badge/pyscn-A-brightgreen)](...)
$ echo $?
0
```

`avg: 1.5` and `high-risk: 0` are computed over `good.py` alone. `complexity.Functions` contains only `tidy` and `<module>`. The unparseable half of the package contributes nothing and costs nothing.

## What the JSON says

The error is recorded correctly:

```json
"Errors": ["[pkg/broken.py] Parse error: syntax errors found in source code"]
```

but `summary.health_score` is `90`, `summary.grade` is `"A"`, and `complexity_score` / `dead_code_score` are both `100`.

A programmatic consumer *can* detect the problem, but only by reading `complexity.Errors` explicitly. The summary block does not help: `complexity.Summary` reports `"FilesAnalyzed": 1` with no corresponding total or skipped count, so the shortfall is only visible to a caller that already knows how many files it passed in.

## Why this matters

The failure mode is silent in exactly the situation the tool is meant to protect: a broken file scores better than a working one, because a file that yields no functions yields no complexity, no dead code, no clones and no coupling. Any workflow that gates on `pyscn check` or on the health score will wave through a commit that does not even import.

I hit this while measuring before/after quality of LLM-generated patches. A model that produced a patch corrupting `django/db/models/query.py` (`return 0 objs`) showed up as the single largest quality *improvement* in the run: 128 functions and 391 total complexity before, 0 and 0 after. Without cross-checking `complexity.Errors` by hand, that reads as a spectacular refactor.

## Suggested fix

Any of these would close the hole; the first two seem most important:

1. Make unparseable files fail the gate. `pyscn check` should exit non-zero (exit code 2, "analysis error", per the documented convention) when any target file failed to parse, unless explicitly waived by a flag such as `--allow-parse-errors`.
2. Withhold or penalise the score. Either refuse to emit a grade when `analyzed_files < total_files`, or subtract a term proportional to the unparsed fraction. Emitting `Grade: A` plus a README badge for a package that does not compile is the part that misleads most.
3. Add `TotalFiles` / `SkippedFiles` alongside the existing `FilesAnalyzed` in `complexity.Summary`, so a consumer can spot the shortfall without having to cross-reference its own input list against `complexity.Errors`.
4. Surface the count in the human summary, e.g. a line like `⚠️ 1 of 2 files skipped (parse errors)` next to the health score, rather than only a warning line above it that scrolls away.

## Environment

- pyscn v1.29.0 (commit 81b7ec0), built from source
- go1.24.6, darwin/arm64

DaisukeYoda:
Fixes #690.

## The problem

When a file fails to parse, pyscn detects it and prints a warning, then drops the file from the analysis entirely. Nothing downstream reflected that: the file yields no functions, no dead code, no clones and no coupling, so it cleared every threshold by contributing nothing at all.

```console
$ python3 -c "import ast; ast.parse(open('pkg/broken.py').read())"
SyntaxError: invalid syntax

$ pyscn check pkg
Warning: Failed to parse file pkg/broken.py: syntax errors found in source code
✅ Code quality check passed
$ echo $?
0
```

`pyscn analyze` scored the same package 90/100, Grade A, and offered a README badge for it. Worse than the issue reported: a directory containing *only* unparseable files also passed, because the "no functions found to analyze" guard is satisfied by raw metrics, which are computed from text before parsing.

The failure mode is silent in exactly the situation the tool is meant to protect. The issue author hit it measuring LLM-generated patches: a patch that corrupted a file (`return 0 objs`) showed up as the single largest quality *improvement* in the run — 128 functions and 391 total complexity before, 0 and 0 after.

## The fix

**1. `check` fails on unanalyzable files.** A new `reportUnanalyzedFiles` helper turns per-file analysis errors into a hard failure across complexity, deadcode, mockdata and DI. `--allow-parse-errors` keeps the old report-only behaviour for callers knowingly analyzing a mixed-syntax tree.

**2. Exit code 2 now exists.** `check --help` has always documented `2: Analysis failed`, but `main` exited 1 for everything. Added `analysisError` plus `exitCodeFor`, so an incomplete analysis is distinguishable from a quality verdict. Also set `SilenceUsage` — dumping the flag list after every failed CI run buried the findings above it.

**3. The health score charges for unanalyzed files.** A penalty proportional to the skipped fraction, with both bounds anchored to the grade thresholds rather than picked by hand:

| Constant | Value | Derivation |
| --- | --- | --- |
| `MinParseErrorPenalty` | 11 | `100 - GradeAThreshold + 1` — any skipped file forfeits an A |
| `MaxParseErrorPenalty` | 56 | `100 - GradeDThreshold + 1` — a wholly unparseable target cannot rank above F |

The floor matters: a purely proportional term would round 1-of-1000 to zero and stay silent. The penalty sits outside the per-category scores, since it belongs to no single analysis.

**4. File accounting is now truthful.** `ComplexitySummary` gains `TotalFiles` and `SkippedFiles`. `calculateSummary` set both `TotalFiles` and `AnalyzedFiles` from `FilesAnalyzed`, which made them always equal and the shortfall structurally undetectable; `skipped_files` was declared in the JSON schema but never assigned anywhere in the repo. The count is also restated next to the health score, where the per-file warnings printed before the run had scrolled away.

## After

```console
$ pyscn check pkg
complexity: [pkg/broken.py] Parse error: syntax errors found in source code
❌ Complexity analysis failed: 1 file(s) could not be analyzed (use --allow-parse-errors to ignore)
Error: analysis failed with errors
$ echo $?
2

$ pyscn analyze pkg --no-open
Health Score: 62/100 (Grade: C)
⚠️  1 of 2 files skipped (parse errors) - excluded from every score below
```

JSON: `total_files: 2, analyzed_files: 1, skipped_files: 1, health_score: 62, grade: "C"`, and `complexity.Summary` reports `FilesAnalyzed: 1, TotalFiles: 2, SkippedFiles: 1`.

## Behaviour change

This is intentionally breaking for anyone whose tree contains Python that pyscn cannot parse — Python 2 sources, generated files, deliberately invalid fixtures. Those runs went from silently passing to exit 2. `--allow-parse-errors` restores the previous behaviour, and health scores on such trees drop by at least 11 points.

## Testing

- `cmd/pyscn/check_parse_errors_test.go` — the gate fails with exit code 2 and names the file; `--allow-parse-errors` waives it while still reporting; a fully parseable package still passes.
- `domain/parse_error_score_test.go` — the penalty table (clean / 1-of-50 / half / none parsed), category scores stay untouched, `Validate` rejects an impossible skipped count.
- `app/analyze_summary_files_test.go` — the summary carries the shortfall instead of flattening it.
- Extended the existing complexity-service subtests to assert the new counts.
- Full suite passes; `go vet` and `gofmt` clean.

Docs updated across all four locales: `cli/check.md`, `integrations/ci-cd.md`, `output/health-score.md`, `output/schemas.md`.

Not addressed: #677 is the sibling problem — parse errors that are never *detected*. This PR assumes detection works and fixes what happens afterwards.


DaisukeYoda:
Addressed all four in b0fd273.

**1. `--select clones` / `--select deps`** — gated clones, qualified the docs for deps.

The clone service already had the errors: it was writing them to `os.Stderr` with `fmt.Fprintf` and discarding them. `CloneResponse` now carries `Errors []string` alongside the existing singular `Error` (whole-run failure), and `check` gates on it. Detection failing on its own stays informational — only an `analysisError` (i.e. unreadable/unparseable input) flips `hasErrors`:

```console
$ pyscn check --select clones pkg
clones: [pkg/broken.py] Parse error: syntax errors found in source code
❌ Clone detection failed: 1 analysis error(s), listed above (use --allow-parse-errors to ignore)
$ echo $?
2
```

`deps` is not gated. The errors aren't merely unreported — `module_analyzer.go:229` discards them with a bare `continue`, and there is no per-file error surface anywhere in `DependencyGraph` → `DependencyAnalysisResult` → `AnalyzeDependenciesOnly`. Plumbing one through is a separate change to the analyzer internals, and arguably a separate bug (deps silently ignores unparseable modules without even a warning). Docs now carry a coverage note naming `deps` as the exception rather than claiming blanket coverage; happy to file a follow-up issue for it.

**2. Exit-code contract leak** — fixed rather than documented away. `SetFlagErrorFunc` wraps flag errors, and `resolveCheckConfig` / `validateSelectedAnalyses` failures are now analysis errors:

| Invocation | Before | After |
| --- | --- | --- |
| `check --no-such-flag .` | 1 | 2 |
| `check --select nonsense .` | 1 | 2 |
| malformed discovered config | 1 | 2 |

**3. Dead-code normalization side effect** — documented in `output/health-score.md` (all four locales), under the dead-code penalty: `TotalFiles` now counts unparsed files, so the normalization factor is slightly larger and the penalty slightly smaller than over analyzed files alone; logarithmic, and dwarfed by the ≥11-point parse-error penalty for the same condition.

**4. `len(errs)` as a file count** — dropped the implicit coupling. The message now counts what it actually has: `"%d analysis error(s), listed above"`.

**One regression caught while testing this:** the `Warning: Failed to parse file …` line that `analyze` printed came from the clone service's `os.Stderr` write. Removing it left `analyze` reporting *how many* files were skipped but not *which*. It now lists them from `complexity.Errors` — the slice that was collected and never read in the first place:

```
Health Score: 62/100 (Grade: C)
⚠️  1 of 2 files skipped (parse errors) - excluded from every score below
    [pkg/broken.py] Parse error: syntax errors found in source code
```

New tests cover the clones gate and both invalid-invocation forms. Full suite, `go vet`, `gofmt` clean.

PR #704, opened 2026-08-09
Build aggregate analysis from one project snapshot

M-Hassan-Raza:
Build one canonical parsed-project snapshot and reuse it across aggregate analyze, check, and MCP execution. File coverage, parse diagnostics, and analyzer failures now come from shared typed contracts, independent of analyzer selection.

Dependency and community analysis now consume one captured module graph without rereading source. Check and MCP preserve partial results while treating incomplete analysis as unhealthy.

Verified with the full test and race suites, vet, golangci-lint, both binary builds, and the strict documentation build. The graph reuse benchmark is roughly 2.9x faster with fewer allocations than reparsing for each consumer.


cmd/pyscn/check_policy.go
```diff
@@ -0,0 +1,184 @@
...
+		SkipCommunities:         true,
+		SkipCommunitiesExplicit: true,
+		SelectAnalysesUsed:      true,
+		MinSeverity:             domain.DeadCodeSeverityCritical,
+	}, paths, app.AnalyzeRequestOverrides{
+		ComplexityEnabled:         domain.BoolPtr(!skipComplexity),
+		DeadCodeEnabled:           domain.BoolPtr(!skipDeadCode),
+		SystemEnabled:             domain.BoolPtr(!skipDependencies),
+		SystemAnalyzeDependencies: domain.BoolPtr(!skipDependencies),
+		SystemAnalyzeArchitecture: domain.BoolPtr(false),
```
DaisukeYoda, line 50: This only forces the analyses on; the graph policy now comes from the project config, where the previous implementation pinned it for the gate. With a relative-import cycle in place, setting `follow_relative = false` in the checked-out repo's own `.pyscn.toml` makes this gate report a clean pass (exit 0) where main fails it (exit 1). A repository can silently disable its own CI gate. Please pin the graph policy here again.

app/analyze_usecase.go
```diff
@@ -294,10 +378,18 @@ func (uc *AnalyzeUseCase) execute(ctx context.Context, useCaseCfg AnalyzeUseCase
...
-		snapshot = service.BuildProjectSnapshotWithOptions(ctx, files, service.ProjectSnapshotOptions{
-			IncludeRawMetrics: uc.complexityUseCase != nil && !useCaseCfg.SkipComplexity,
+	snapshot := service.BuildProjectSnapshotWithOptions(ctx, files, service.ProjectSnapshotOptions{
+		IncludeRawMetrics: uc.complexityUseCase != nil && !useCaseCfg.SkipComplexity,
+	})
+
+	var moduleGraph *service.ProjectModuleGraph
+	var moduleGraphErr error
+	if uc.needsModuleGraph(useCaseCfg) {
+		moduleGraph, moduleGraphErr = snapshot.BuildDependencyGraph(ctx, &service.ModuleGraphOptions{
```
DaisukeYoda: The graph is now built from the aggregate snapshot, which uses the source-file pattern set (`**/*.py`). Dependency analysis is contractually supposed to use the module pattern set, which also covers `.pyi`. A cycle passing through a stub-only module is detected on main but missed here, and only reappears if the user adds `**/*.pyi` by hand.

app/clone_usecase.go
```diff
@@ -152,6 +166,30 @@ func (uc *CloneUseCase) ExecuteAndReturn(ctx context.Context, req domain.CloneRe
...
+func (uc *CloneUseCase) analyzeSnapshotRequest(ctx context.Context, snapshot *svc.ProjectSnapshot, req domain.CloneRequest) (*domain.CloneResponse, error) {
+	if snapshot == nil {
+		return nil, fmt.Errorf("project snapshot is required")
+	}
+
+	finalReq, err := uc.loadAndMergeConfig(req)
+	if err != nil {
+		return nil, domain.NewConfigError("failed to load configuration", err)
+	}
+	finalReq.Paths = snapshot.Paths()
```
DaisukeYoda, line 178: This merges the clone config and then replaces the paths with every snapshot file, so the clone-specific include/exclude patterns are dropped. On a 121-file project with no pyscn config the whole test tree got pulled in: 79 to 115 files, 172 to 302 pairs, duplication 11.6% to 14.8%, health score 54 to 52. Clone detection is quadratic, so this also took `pyscn analyze` from 8.6s to 18.9s, which more than cancels the gains everywhere else.

internal/analyzer/module_analyzer.go
```diff
@@ -497,14 +590,57 @@ func (ma *ModuleAnalyzer) importsFromNode(node *parser.Node) []*ImportInfo {
...
+			candidate := strings.Join(append(append([]string(nil), packageParts[:depth]...), moduleName), ".")
+			if graph.GetModule(candidate) != nil && ma.importMatchesResolvedModule(moduleName, candidate) {
+				return candidate
+			}
+		}
+	}
+	if graph.GetModule(moduleName) != nil {
+		return moduleName
+	}
+	for _, candidate := range graph.GetModuleNames() {
```
DaisukeYoda: This accepts any module whose dotted name ends with the imported segment, which binds third-party imports to same-named local packages. On piccolo_api, `from fastapi import ...` resolves to the project's own `piccolo_api.fastapi`, adding four phantom internal edges and moving the architecture score and community modularity. It also runs on the non-snapshot path, so standalone dependency analysis changes too.

service/complexity_service.go
```diff
@@ -106,12 +107,15 @@ func (s *ComplexityServiceImpl) AnalyzeSnapshot(ctx context.Context, snapshot *P
...
 	filesSkipped := 0
 
-	for _, file := range snapshot.Files {
+	for _, file := range snapshot.files {
 		select {
 		case <-ctx.Done():
 			return nil, fmt.Errorf("complexity analysis cancelled: %w", ctx.Err())
 		default:
 		}
+		if !file.Parsed() {
```
DaisukeYoda: When every file in scope fails to parse, skipping them all leaves no functions and no raw metrics, so the analyzer returns a hard failure instead of an empty result plus diagnostics. `--allow-parse-errors` then stops waiving anything (exit 2 here, exit 0 on main). Mixed trees are fine, but the all-unparseable case is exactly what the flag is for.

mcp/README.md
```diff
@@ -442,16 +442,19 @@ What's the health score of my codebase?
   "health_score": 85,
   "grade": "A",
   "is_healthy": true,
-  "category_scores": {
+  "partial": false,
```
DaisukeYoda, line 456: The documented shape changed but `HandleGetHealthScore` didn't, so these fields are never emitted, `architecture_score` is dropped from the example although the tool does return it, and that handler still discards partial results unlike `HandleAnalyzeCode` which was fixed in this PR.

service/project_snapshot.go
```diff
@@ -91,10 +127,11 @@ func BuildProjectSnapshotWithOptions(ctx context.Context, paths []string, option
 	wg.Wait()
 
 	for idx, path := range paths {
-		if snapshot.Files[idx] == nil {
-			snapshot.Files[idx] = cancelledProjectFile(path, ctx.Err())
+		if snapshot.files[idx] == nil {
+			snapshot.files[idx] = cancelledProjectFile(path, ctx.Err())
 		}
 	}
+	snapshot.Files = projectFileProjections(snapshot.files)
```
DaisukeYoda: Nothing in the codebase reads `Files`; it's a full deep copy of every AST and source buffer. It's about a quarter of snapshot build time and a third of its allocations, and it doubles the retained memory of the parsed project. An exported field documented as not affecting the state behind it is also easy to misuse; a method would be clearer.

DaisukeYoda (changes requested):
Great work on this. The typed coverage and failure contracts are a real improvement, and the snapshot reuse pays off: five of the six check gates got 2-3x faster on my test project.

That said, I ran into a few practical problems while testing it, so please have a look at the inline comments. The dependency gate one is the important one.

Given the size, splitting the snapshot plumbing from the behavior changes would also make this much easier to verify.

M-Hassan-Raza:
@DaisukeYoda please check again — should be solid now. Thanks!

internal/analyzer/module_analyzer.go
```diff
@@ -497,14 +590,41 @@ func (ma *ModuleAnalyzer) importsFromNode(node *parser.Node) []*ImportInfo {
...
 // resolveImport resolves an import to a module name
-func (ma *ModuleAnalyzer) resolveImport(imp *ImportInfo, fromFile string) string {
+func (ma *ModuleAnalyzer) resolveImport(graph *DependencyGraph, imp *ImportInfo, fromFile string, capturedOnly bool) string {
 	if imp.IsRelative {
 		return ma.resolveRelativeImport(imp, fromFile)
 	}
+	if resolved := ma.resolveAbsoluteImportFromGraph(graph, imp); resolved != "" {
+		return resolved
+	}
+	if capturedOnly {
```
DaisukeYoda, line 600: **`capturedOnly` resolution drops same-directory (script-style) imports, silently erasing dependency edges**

First, the intent here makes sense to me: `resolveAbsoluteImportWithProject` stats the filesystem for every import candidate (`resolveModuleFile` / `resolvePackageInit`), which violates the "captured snapshot never touches disk" invariant this PR establishes, and the tightening to exact matching is clearly aimed at the false positive locked in by `TestModuleAnalyzer_AbsoluteThirdPartyImportDoesNotBindByLocalSuffix` (local `piccolo_api/fastapi.py` vs. the real `fastapi`).

The issue is that exact graph-name matching is stricter than what the old path actually did, and the difference shows up as silent false negatives in the quality gate.

Repro:

```
sub/ma.py:  import mb
sub/mb.py:  import ma
```

- `main`: `circular dependency detected: sub.ma -> sub.mb`, exit 1
- this PR: `✅ Code quality check passed`, exit 0

On this repo's own `testdata/python`, `pyscn check --select deps .` drops from 10 cycles to 1, and `pyscn analyze`'s dependency score goes 25/100 → 80/100. Since dependency and community analysis both consume the captured graph now, this affects `check --select deps`, `analyze`, community detection, and MCP `analyze_code`.

Why the new test doesn't catch it: what protected the old path in the piccolo case wasn't strictness, it was *search distance*. `resolveAbsoluteImportWithProject` searched only the importing file's directory and its parent — neither `piccolo_api/routers/admin/` nor `piccolo_api/routers/` contains `fastapi.py`, so `importMatchesResolvedModule`'s suffix rule never got a chance to bind it. For `sub/ma.py: import mb` the sibling *is* in the current directory, so the old path resolved it to `sub.mb`.

Suggestion that keeps the no-disk-access property: when the exact graph lookup misses, retry against the graph using the importing module's package prefix, walking up at most two levels (mirroring the old `currentDir` + parent search):

- `sub.ma` imports `mb` → try `sub.mb` → hit ✅
- `piccolo_api.routers.admin.api` imports `fastapi` → `piccolo_api.routers.admin.fastapi` ✗ → `piccolo_api.routers.fastapi` ✗ → stop → stays external ✅ (the new test stays green)

Those are map lookups, not stats, so both the snapshot invariant and the ~2.9x graph-reuse win are preserved.

Separately, in this same block: `resolveAbsoluteImport` returns `""` for a stdlib module when `includeStdLib` is false *without* falling through to the `includeThirdParty` branch, but here `includeThirdParty` is checked unconditionally afterwards. With the default config (`include_stdlib=false`, `include_third_party=true`), `import os` therefore classifies as third-party. It's masked today because `DependencyGraph.AddDependency` drops edges whose target node doesn't exist, but it will produce bogus edges the moment external nodes are materialized.

DaisukeYoda: @M-Hassan-Raza Could you check this point? (There might be better implementation)

DaisukeYoda:
@M-Hassan-Raza Thank you for updating. Let me know if I can review!

M-Hassan-Raza:
@DaisukeYoda Please take a look now! Sorry for the delay I was super occupied with job related matters :)

DaisukeYoda:
@M-Hassan-Raza Could you check the below points as well? (and some conflicts) Thanks!

**Likely regressions (blocking):**

1. **mockdata/DI ignore request-level patterns** — `MockDataServiceImpl.AnalyzeSnapshot` (service/mock_data_service.go:127) and `DIAntipatternServiceImpl.AnalyzeSnapshot` (di_antipattern_service.go:95) iterate the shared snapshot without consulting the `IncludePatterns`/`ExcludePatterns` merged in the usecase. Files excluded by the snapshot defaults (e.g. `**/migrations/**`) silently stop being checked — findings that failed the gate on main now pass.
2. **Clone pattern base mismatch** — `selectedAnalysisProjectFiles` (service/project_snapshot.go:304) matches clone include/exclude patterns against project-root-relative paths, while collection matched walk paths. Anchored patterns like `app/**/*.py` can silently select 0 files when running from a subdirectory.
3. **Gate can pass on a failed run** — `isInformationalCloneAnalysisError` (cmd/pyscn/check.go:268) classifies by scanning `response.Failures`, so a real `build analysis response` error co-occurring with a clone-only failure is downgraded to informational → exit 0.
4. **Dead error-recording code** — the new `if !file.Parsed() { continue }` guard in every `AnalyzeSnapshot` (e.g. cbo_service.go:109) makes the ReadErr/ParseErr recording blocks below unreachable, so per-analyzer errors diverge between standalone commands and `analyze`.

**Intentional but incomplete:**

5. **`--allow-parse-errors` now waives only parse diagnostics** (per "waive only parse diagnostics") — but the flag help (check.go:129) and website/docs/cli/check.md in all 4 locales still promise the old behavior. Needs doc updates + a changelog note as a breaking change.
6. **Coverage now counts `.pyi` stubs** — intentional for the module graph, but a single unparsable stub (a file no quality analyzer reads) now fails `check` with no waiver, and total_files inflates (the #690 contract test was deleted). Consider excluding stub-only diagnostics from the gate.
7. **Bare include patterns now match nested basenames** (`main*.py` matches `pkg/main_utils.py`) — test was flipped deliberately, but this user-visible change is undocumented.

M-Hassan-Raza:
@DaisukeYoda please check now? Thanks!

DaisukeYoda (approved):
Thanks!!!

PR #709, opened 2026-08-12
Report class-body complexity in its own scope

### LINKED ISSUE #694: [BUG] Class-body control flow is counted by cyclomatic complexity but by no cognitive complexity scope
## Describe the bug

Control flow written directly in a **class body** (not inside a method) is counted by cyclomatic complexity but by **no cognitive complexity score at all**. It is silently dropped.

The cause is a mismatch between two layers:

- `CFGBuilder.buildClass` (`internal/analyzer/cfg_builder.go`) processes a class body into the **enclosing** CFG. Only its methods get their own CFG via `buildNestedFunction`; a class itself never gets a CFG entry in `BuildAll`'s `functionCFGs`.
- `traverseForCognitive` (`internal/analyzer/cognitive_complexity.go`) treats `NodeClassDef` as a scope boundary and returns immediately, so the class body is not counted toward the enclosing scope either.

Result: the class body's control flow inflates the enclosing scope's **cyclomatic** complexity while contributing 0 to **cognitive** complexity, and no separate `Config` scope exists to carry it. The two metrics for `<module>` disagree in a way a user cannot explain from the source.

Note this is not a regression introduced by #661. `collectASTStatementMetrics` (`internal/analyzer/complexity.go:247`) has always returned early on `ClassDef`, so `IfStatements`/`LoopStatements` already excluded class bodies. #661 made cognitive complexity consistent with those counters, which is correct — but it also makes the remaining divergence against cyclomatic complexity visible.

## To Reproduce

1. Save the file below as `b.py`
2. Build CFGs with `CFGBuilder.BuildAll` and run `CalculateComplexity` on each (equivalently, `pyscn analyze b.py` and inspect the `<module>` row)

## Expected behavior

Either:

- **(a)** the class body's control flow is attributed to a `Config` scope of its own (a class gets its own CFG, symmetric to how nested functions are handled), or
- **(b)** it is attributed to the enclosing scope by *both* metrics.

Option (a) matches the scope model that #661 established and is likely the better fit, but it changes the set of reported scopes, so it needs a deliberate decision.

Whichever is chosen, cyclomatic and cognitive complexity must not disagree about whether the code exists.

## Example Code

```python
import sys

class Config:
    if sys.version_info >= (3, 11):
        BACKEND = "fast"
    elif sys.version_info >= (3, 9):
        BACKEND = "medium"
    else:
        BACKEND = "slow"

    for name in ("a", "b", "c"):
        locals()[name.upper()] = name
```

This is a real-world pattern (version gating, `TYPE_CHECKING` branches, generated class attributes).

## Output

Measured on `fix/issue-658-cognitive-complexity-crosses-scope-bou` @ `867e789`:

```
<module>             cyclo=4 cognitive=0 ifs=0
```

`cyclo=4` counts the `if` + `elif` + `for` (+1 base). `cognitive=0` and `ifs=0` count none of them. There is no `Config` entry in the CFG map — `<module>` is the only scope reported.

## Environment

- OS: macOS (Darwin 24.5.0)
- Go Version: as pinned in `go.mod`
- pyscn Version: branch `fix/issue-658-cognitive-complexity-crosses-scope-bou` @ `867e789` (post-#661)

## Additional context

Found while reviewing #661. Deliberately left out of that PR because fixing it requires changing CFG construction (`buildClass`), which is outside the scope of a cognitive-complexity traversal fix.

Related: #658, #661

M-Hassan-Raza:
Fixes #694.

Give executable class suites their own typed CFG and complexity result. Class-body branches no longer inflate the enclosing module; cyclomatic, cognitive, statement, and nesting metrics now share one owner.

Keep function aggregates stable, expose class scopes separately across CLI, JSON, YAML, CSV, HTML, and MCP, and include non-dilutable class hotspots in health and complexity gates.

Verified with full, race, shuffled, and repeated scope tests, lint, strict docs, coverage, and CLI/MCP builds.


service/output_formatter.go
```diff
@@ -160,6 +137,39 @@ func (f *OutputFormatterImpl) formatText(response *domain.ComplexityResponse) (s
...
+		return
+	}
+
+	builder.WriteString(utils.FormatSectionHeader(title))
+	builder.WriteString(utils.FormatTableHeader("Scope", "Complexity", "Cognitive", "SLOC", "Risk"))
+
+	for _, scope := range scopes {
+		// Convert domain risk level to standard risk level
+		var standardRisk RiskLevel
+		switch scope.RiskLevel {
```
DaisukeYoda, line 150: `scope.RiskLevel` values are lowercase (`"high"` etc.), so these capitalized cases never match and every row falls through to `RiskLow` — including the new class-scope table. Compare against `domain.RiskLevelHigh/Medium/Low` instead.

service/dead_code_service.go
```diff
@@ -243,19 +243,23 @@ func (s *DeadCodeServiceImpl) analyzeProjectFile(file *ProjectFile, req domain.D
...
 	totalFindings := 0
 	unfilteredFindings := 0
 	affectedFunctions := 0
 
-	for functionName, cfg := range cfgs {
+	functionCount := 0
+	for _, scopedCFG := range cfgs {
 		// Skip the main module CFG for now, focus on functions
-		if functionName == domain.ModuleFunctionName {
+		if scopedCFG.Scope.Kind != domain.AnalysisScopeFunction {
```
DaisukeYoda: Skipping every non-function CFG drops dead-code detection inside class suites entirely (same skip in `internal/analyzer/dead_code.go:178`). Before this PR a class body was folded into its enclosing function's CFG, so unreachable code after e.g. a `raise` in the class body was reported; now it is never analyzed anywhere, and `pyscn check` passes on code that previously failed. Should class-suite CFGs be included here?

cmd/pyscn/check.go
```diff
@@ -474,16 +474,20 @@ func (c *CheckCommand) checkComplexity(cmd *cobra.Command, args []string) (int,
...
 	}
 
-	// Count functions that exceed the maximum complexity threshold, plus
-	// functions that are too long. Length is orthogonal to McCabe, so a flat
+	// Count scopes that exceed the maximum complexity threshold, plus functions
+	// that are too long. Length is orthogonal to McCabe, so a flat
 	// 200-line function is an issue on its own and both can fire at once.
 	issueCount := 0
-	for _, function := range response.Functions {
+	for _, function := range response.ReportedScopesByComplexity() {
```
DaisukeYoda: The gate iterates `ReportedScopesByComplexity()`, which is post-`min_complexity`/`report_unchanged` filtering. With `min_complexity = 12`, a CC-11 scope (over the default gate of 10) is filtered out of the response and `pyscn check` exits 0. Shouldn't this use `AnalyzedScopesByComplexity()`, which this PR added for exactly this purpose? `docs/algorithms/complexity.md` (in this PR) says gates use the complete pre-filter population, and the issue counting in `mcp/handlers.go` has the same problem.

internal/analyzer/cfg_builder.go
```diff
@@ -307,13 +308,16 @@ func (b *CFGBuilder) processStatement(stmt *parser.Node) {
...
+		if err := b.buildNestedScope(stmt); err != nil {
 			b.logError("error in nested function: %v", err)
 		}
 
 	case parser.NodeClassDef:
-		// Build separate scope for nested classes
-		b.buildClass(stmt)
+		// A class suite is its own execution scope. Keep only the definition
+		// binding and header evaluation in the enclosing graph.
+		if err := b.buildNestedScope(stmt); err != nil {
```
DaisukeYoda: Unreviewed side effect on clone detection: `semantic_similarity.go` builds fragment CFGs with this same builder, so a function fragment containing a class definition now loses the class-body control flow and can fall below the semantic min-cyclomatic gate — previously detected Type-4 clone pairs silently disappear. If intended, worth pinning with a test; otherwise needs handling.

domain/analyze.go
```diff
@@ -380,8 +387,11 @@ func (s *AnalyzeSummary) calculateComplexityPenalty() int {
 	mccabePenalty := coredomain.LinearPenalty(s.AverageComplexity, 2.0, 15.0)
 	cognitivePenalty := coredomain.LinearPenalty(s.AverageCognitiveComplexity, 15.0, float64(DefaultCognitiveComplexityThreshold))
 	nestingPenalty := coredomain.LinearPenalty(s.AverageNestingDepth, 3.0, float64(DefaultNestingDepthThreshold))
+	classMcCabePenalty := coredomain.LinearPenalty(float64(s.MaxClassComplexity), 2.0, 15.0)
```
DaisukeYoda: Two concerns:
1. The start/max values (2.0/15.0 McCabe, 3.0 nesting) are calibrated for population *averages* but are applied here to a single-scope *maximum*. A class with `if TYPE_CHECKING:` plus a couple of version checks (suite CC 4-5) already takes a penalty, and one CC-15 class zeroes the complexity score, while a function needs a project-wide average of 15 for the same effect.
2. The bare literals are now duplicated across the function and class variants — consider named constants, as the cognitive variant already does with `DefaultCognitiveComplexityThreshold`.

mcp/handlers.go
```diff
@@ -802,26 +810,30 @@ func formatComplexitySummary(result *domain.ComplexityResponse, threshold int, m
...
-			"total_functions":    result.Summary.TotalFunctions,
-			"functions_parsed":   result.Summary.FunctionsParsed,
-			"max_complexity":     result.Summary.MaxComplexity,
-			"average_complexity": result.Summary.AverageComplexity,
-			"threshold":          threshold,
+			"total_issues":             totalIssues,
+			"total_functions":          result.Summary.TotalFunctions,
+			"total_class_scopes":       result.Summary.TotalClassScopes,
+			"functions_parsed":         result.Summary.FunctionsParsed,
+			"max_complexity":           maxComplexity,
```
DaisukeYoda: This silently changes the meaning of the existing `max_complexity` field (function max → max across all scopes, with the old value renamed to `max_function_complexity`), which will surprise clients tracking it over time. Consider keeping `max_complexity` as-is and adding a new field, or at least calling the change out. Also, `mcp/README.md` still documents `max_complexity` as "0 = no limit", contradicting the updated `tools.go` description ("0 uses the default (10)").

domain/complexity.go
```diff
@@ -132,6 +149,20 @@ type FunctionComplexity struct {
 	RiskLevel RiskLevel
 }
 
+// ResolvedScopeKind preserves the public struct's historical zero-value
+// behavior. Results produced by the analyzer always carry an explicit kind;
+// externally constructed legacy records remain functions except for the
+// established module pseudo-record.
+func (f FunctionComplexity) ResolvedScopeKind() AnalysisScopeKind {
```
DaisukeYoda: This silent fallback (and the one in `AnalyzedScopesByComplexity()`) masks construction bugs: a caller that forgets to set `ScopeKind` gets its records silently classified as functions, and a response missing the `Analyzed*` populations silently degrades to the filtered reported collections. Could these fail fast instead?

cmd/pyscn/check.go
```diff
@@ -474,16 +474,20 @@ func (c *CheckCommand) checkComplexity(cmd *cobra.Command, args []string) (int,
...
+		name := function.Name
+		if function.ResolvedScopeKind() == domain.AnalysisScopeClass {
+			name = "class scope " + name
```
DaisukeYoda: This label formatting duplicates `complexityScopeLabel()` added in `mcp/handlers.go` in this same PR, and tests in both layers assert the literal string, so the two copies must stay in lockstep. Consider moving the helper to `domain` (next to `ResolvedScopeKind`) and calling it from both.

mcp/handlers.go
```diff
@@ -783,16 +789,18 @@ func formatComplexitySummary(result *domain.ComplexityResponse, threshold int, m
 		threshold = 10
 	}
 
-	// Filter functions that exceed threshold
-	for _, fn := range result.Functions {
+	// Filter all executable scopes that exceed the function complexity gate.
+	scopes := result.ReportedScopesByComplexity()
+	maxComplexity, averageScopeComplexity := summarizeScopeComplexity(result.AnalyzedScopesByComplexity())
```
DaisukeYoda: `summarizeScopeComplexity` only computes max and average, so sorting the full analyzed population (here and again in `formatComplexityDetailed`) is unnecessary — a single unsorted pass over `r.AnalyzedFunctions` + `r.AnalyzedClassScopes` would do, and would also remove the duplicated two-line block.

M-Hassan-Raza:
@DaisukeYoda Please take a look now! Sorry for the delay I was super occupied with job related matters :)

mcp/handlers.go
```diff
@@ -774,94 +784,128 @@ func (h *HandlerSet) HandleGetHealthScore(ctx context.Context, request mcp.CallT
...
+
+type complexityGateAnalysis struct {
+	Issues            []domain.FunctionComplexity
+	TotalIssues       int
+	MaxComplexity     int
+	AverageComplexity float64
+}
+
+func analyzeComplexityGate(result *domain.ComplexityResponse, threshold int, maxResults int) (complexityGateAnalysis, error) {
+	scopes, err := result.AnalyzedScopes()
```
DaisukeYoda: Using the pre-filter `AnalyzedScopes()` here fixes the gate bypass, but it changes the tool contract in two ways:

1. `min_complexity` no longer affects `issues`/`total_issues` in summary/detailed modes — every scope over `max_complexity` is counted regardless, so the same arguments now return a much longer issue list than before.
2. `full` mode still serializes the *filtered* `Functions`/`ClassScopes`, so `total_issues` counts scopes that the returned records omit — the two can't be reconciled.

Either apply `min_complexity` to the gate's issue list, or make `full` mode serialize the same population the gate counts.

service/output_formatter.go
```diff
@@ -179,15 +188,15 @@ func (f *OutputFormatterImpl) formatCSV(response *domain.ComplexityResponse) (st
...
+	header := []string{"Function", "Complexity", "Cognitive Complexity", "Risk", "Nodes", "Edges", "Nesting Depth", "If Statements", "Loop Statements", "Exception Handlers", "SLOC", "Scope Kind"}
 	if err := writer.Write(header); err != nil {
 		return "", domain.NewOutputError("failed to write CSV header", err)
 	}
 
-	// Write data rows
-	for _, function := range response.Functions {
+	// Write all reported executable scopes. The appended kind column keeps the
+	// established function columns stable while making class rows unambiguous.
+	for _, function := range response.ReportedScopes() {
```
DaisukeYoda: `ReportedScopes()` concatenates the independently sorted `Functions` and `ClassScopes` lists, so the CSV rows no longer follow the requested `--sort` order across the whole file. With `sort_by=complexity` you can get functions `20, 12, 5` followed by class scopes `30, 8` — a consumer reading top-down picks 20 as the worst scope instead of 30.

`generateComplexitySuggestions` (domain/suggestion.go:84) consumes the same concatenation and has the same discontinuity. Consider merging the two lists under the active sort key before emitting.

DaisukeYoda:
@M-Hassan-Raza No worries at all! Thanks for coming back. 

Could you check the above two points before merging?

M-Hassan-Raza:
> @M-Hassan-Raza No worries at all! Thanks for coming back.
> 
> Could you check the above two points before merging?

Thanks for taking a look! pls check now? Should be good to go (I think)

DaisukeYoda (approved):
@M-Hassan-Raza LGTM! Many thanks!

PR #720, opened 2026-08-20
Exclude method-call targets from LCOM4 InstanceVariables count

### LINKED ISSUE #678: [BUG][auto] lcom4-instance-vars-count-method-calls: InstanceVariables counts intra-class method call targets
## Summary

`LCOMResult.InstanceVariables` counts intra-class method call targets as instance variables. A class with two attributes and one `self.helper()` call reports `InstanceVariables: 3`.

The field is documented as "Distinct `self.xxx` variables" (`internal/analyzer/lcom.go:27`), and method names are not variables. `LCOM4`, `MethodGroups`, and the risk level are all unaffected — this is a wrong auxiliary count, not a wrong cohesion verdict.

## Repro

```python
class C:
    def __init__(self):
        self._a = 1
        self._b = 2

    def helper(self):
        return self._a

    def run(self):
        self.helper()
        return self._b
```

```console
$ pyscn analyze --json --no-open .
```

## Expected

`InstanceVariables: 2` — the class declares exactly `_a` and `_b`.

## Actual output

From the generated `analyze_*.json`:

```json
{
  "Name": "C",
  "Metrics": {
    "LCOM4": 1,
    "TotalMethods": 3,
    "ExcludedMethods": 0,
    "InstanceVariables": 3,
    "MethodGroups": [["__init__", "helper", "run"]]
  }
}
```

`3` is `_a` + `_b` + `helper`. The grouping is right; only the count is inflated.

## Likely cause

In `internal/analyzer/lcom.go`, `collectMethods` returns per-method variable sets and `methodCalls` separately, but the `self.X` accesses that were resolved as intra-class method calls are still present in the variable sets that feed `allVars`:

```go
// internal/analyzer/lcom.go:108-118
allVars := make(map[string]bool)
for _, vars := range methods {
    for v := range vars {
        allVars[v] = true
    }
}
for v := range declaredFields {
    allVars[v] = true
}
result.InstanceVariables = len(allVars)
```

Names recorded in `methodCalls` (or matching a method defined in the class body) look like they should be excluded before the count.

## Not covered by existing tests

The nearest existing assertion is `lcom_test.go:887` (`assert.Equal(t, 3, r.InstanceVariables, "self.x, self.y, self.foo")`), but in that fixture `self.foo` is a genuine attribute assignment (`def ignored(self): self.foo = 'bar'`), so it does not sanction counting call targets.

## Impact in the wild

Found while auditing `gi0baro/tonio@269e38a`. `_SignalReceiver` (`tonio/_signals.py:9`) reports:

```json
{"LCOM4": 4, "TotalMethods": 7, "ExcludedMethods": 0, "InstanceVariables": 6}
```

but the class declares exactly four: `__slots__ = ['_sigs', '_chr', '_chw', '_inner']`. The extra two are `self._init_channel` and `self._register_coros`, both called from `__enter__`.

## Priority rationale

P2 per the rubric: visible but bounded inaccuracy. A reported field is internally inconsistent with the source, but the cohesion result the user acts on (`LCOM4`, `MethodGroups`, risk level) is correct.

## Version

`pyscn v1.26.4-9-g37793ce`

Found via the FP-audit skill in repo `gi0baro/tonio@269e38af7a7d6099e0b68b3c3e70024d56123951`.

zebra-jay:
## Summary
Fixes a bug in LCOM4's `InstanceVariables` metric where method-call
targets (e.g. `helper` in `self.helper()`) were being counted as
instance variables alongside genuine `self.x` attribute accesses,
inflating the reported count.

## Type of Change

## Related Issues
Closes #678

## Changes
- `extractInstanceVars`'s matched attribute nodes weren't being
  checked against `methodCalls` before being added to a method's
  variable set, unlike the existing reclassification already done
  for `@property` reads.
- Added a reclassification step after `extractMethodCalls` runs:
  any name in `vars` that's also a call target is removed, same mechanism
  as the property-name handling already in place.

## Testing
- [ ] `make test` passes locally
- [ ] `make lint` passes locally
- [ ] Manual testing completed

Added two cases to `TestLCOMAnalyzer_MethodAcessCountsAsInstanceVariable`:
- The issue's own repro (`C` class), expecting `InstanceVariables: 2`.
- A real-world class from the issue's audit target (`gi0baro/tonio`'s
  `_SignalReceiver`), expecting `InstanceVariables: 4`. Note: current
  `main` reports `7` for this class rather than the issue's originally
  reported `6` — as another call-target method (`call_coros`) has been
  added. This fix resolves all three. 

## Documentation
- [ ] Updated godoc comments

## Additional Context
**Contributor context**
- First-time contributor to this repo.
- Used an LLM as a learning aid while tracing the bug and navigating
  Go for the first time; the diagnosis, fix, and tests are mine.

**Verification scope**
- Ran `go test ./internal/analyzer/... -v` — passes.
- Ran `gofmt -l internal/analyzer/` and `go vet ./internal/analyzer/...`
  — both clean. This is the only package the change touches.
- Did not run the full-repo `make test`/`make lint` (`go test ./...`):
  it includes a `service` package test that ran past the default
  10-minute timeout on this machine and got killed by Go's watchdog,
  unrelated to this change. Can rerun with a longer timeout if needed.


DaisukeYoda:
Hi @zebra-jay, thanks for the contribution and the clear PR description. Welcome!

One issue with the approach: `vars` also feeds the LCOM4 graph's shared-variable edges, so deleting every call target changes the graph, not just the count. Verified regressions vs `main`: stored callables (`self._cb = cb` / `self._cb()`), calls to `@abstractmethod`/static/inherited methods, and self-recursion all now split cohesive classes (LCOM4 1 -> 2). The constructor branch (lines 202-217) also doesn't get the filter, so #678 still reproduces for helper calls in `__init__`.

Suggestion: instead of subtracting after the fact, gate the deletion on actual class methods. `FindClassMethods` (di_helpers.go) already enumerates them. That fixes all cases at once, in both branches, and your tests carry straight over.

Nits: remove the commented-out `delete(vars, name)`; the `if vars[method]` guard is redundant (`delete` is a no-op on missing keys); the test name has a typo ("Acess") and asserts the opposite of what it says.

Happy to answer questions on the LCOM4 internals. Thanks again!

DaisukeYoda:
@zebra-jay Any luck?

Issue #706, opened 2026-08-12
Follow-up cleanups from PR #697 review (fail-fast, dead reported/parsed paths, duplication)


DaisukeYoda:
Non-urgent follow-ups from the review of PR #697 (display-only complexity filters). Roughly in priority order:

1. **Fail-fast when `AnalyzedFunctions` is unset** — `attachDirectoryComplexity` (`app/complexity_usecase.go:262`) silently produces an empty `ByDirectory` when the non-serialized `AnalyzedFunctions` field is not populated. Nothing enforces the invariant at compile time, so a future producer that only sets `Functions` makes the Directory Complexity section vanish with no error. Return an explicit error instead of a silent empty result.

2. **Restore a "reported / parsed" indicator** — summaries now show the complete population (`TotalFunctions`) while the function list shows the filtered subset, with no signal that filtering hid functions. E.g. `pyscn complexity --min-complexity 5`: summary says "Total Functions: 105" while the table lists 29 rows, and the old "29 reported / 105 parsed" line can no longer fire because `FunctionsParsed` is now always set equal to `TotalFunctions` (`service/complexity_service.go:437-438`). Affects `service/output_formatter.go:65,342` and `service/analyze_formatter.go:76`.

3. **Remove dead reported/parsed branches** — with `FunctionsParsed == TotalFunctions` always, `formatFunctionCoverage`'s `parsed != reported` branch (`service/output_formatter.go:383`) and the HTML template conditional (`service/analyze_formatter.go:1029-1030`) are unreachable.

4. **Deduplicate response construction** — `Analyze` and `AnalyzeSnapshot` (`service/complexity_service.go` ~73-88 and ~143-160) contain an identical block (module rollup, filter, sort, `generateSummary`, response construction) that PR #697 had to edit in lockstep. Extract a shared `buildComplexityResponse` helper.

5. **Drop the discarded return value** — `filterFunctions` (`service/complexity_service.go:299`) still computes and returns `functionsParsed`, which every caller now discards with `_`, and its doc comment describes the obsolete contract.

6. **`ComplexitySummary.FunctionsParsed` is redundant** — permanently equal to `TotalFunctions` for every producer. If JSON/YAML schema compatibility requires keeping it, mark it deprecated in the comment; otherwise remove it (also emitted at `mcp/handlers.go:126/807/859`).

Items 2/3/6 interact: decide first whether the reported/parsed distinction should be restored (option 2) or fully removed (options 3+6).

PR #721, opened 2026-08-24
fix: reject missing complexity function population

sanmaxdev:
## Summary
- Reject complexity responses that omit the complete function population before directory rollups.
- Preserve valid empty populations for files that only produce raw metrics.

## Type of Change
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues
Refs #706

## Changes
- Initialize service responses with a non-nil complete function collection.
- Return an explicit invariant error when the complete collection is missing.
- Cover missing and valid empty populations with regression tests.

## Testing
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run --allow-parallel-runners --timeout=10m`
- [x] `go build ./cmd/pyscn`

## Documentation
- [ ] Updated godoc comments
- [ ] Updated README or other docs (not needed)

### DIFF
--- a/app/complexity_usecase.go
+++ b/app/complexity_usecase.go
@@ -259,6 +259,12 @@ func (uc *ComplexityUseCase) AnalyzeFile(ctx context.Context, filePath string, r
 }
 
 func attachDirectoryComplexity(response *domain.ComplexityResponse, projectRoot string) error {
+	if response == nil {
+		return fmt.Errorf("complexity response is required")
+	}
+	if response.AnalyzedFunctions == nil {
+		return fmt.Errorf("complete analyzed function population is required")
+	}
 	byDirectory, err := aggregateComplexityByDirectory(response.AnalyzedFunctions, projectRoot)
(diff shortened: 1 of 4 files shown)


DaisukeYoda:
@sanmaxdev 

Thanks! The fail-fast guards in `attachDirectoryComplexity` are consistent with both producers now initializing `allFunctions` via `make(..., 0)`, all call sites propagate the new error, and the targeted tests pass on this branch.

One minor, non-blocking note: the non-nil `AnalyzedFunctions` invariant is currently enforced at each construction site (`Analyze` / `AnalyzeSnapshot` duplicate the same response literal). Extracting a shared `ComplexityResponse` constructor that guarantees the field would prevent a future producer from silently reintroducing `nil`. Fine as a follow-up.

DaisukeYoda:
@sanmaxdev Any luck?

PR #726, opened 2026-09-01
fix: report unclamped code duplication percentage

### LINKED ISSUE #724: [ENH][auto] duplication-percentage-clamped-at-scoring-threshold: reported code_duplication_percentage is capped at 30% so heavily-duplicated projects are indistinguishable
**Track**: tuning (reported-metric accuracy)

## Summary

`code_duplication_percentage` is clamped to the scoring saturation constant before being stored, so it stops being a measurement. Any project whose true fragment-clone ratio exceeds 30% reports **exactly `30.0`** — indistinguishable from any other such project, whether the real figure is 31% or 95%.

The clamp is **redundant for scoring**: `DuplicationPenalty` already saturates internally. Removing it changes no score and only fixes the reported number.

## Evidence

On `jodal/pykka@30af213` (21 files analyzed), the clone statistics and the summary disagree:

```json
"statistics": { "total_fragments": 61, "total_clones": 32 },
"summary":    { "code_duplication_percentage": 30, "duplication_score": 0 }
```

True ratio is `32 / 61 = 52.46%`, reported as `30.0`. The terminal summary prints `8.6%`-style text from this same field:

```
Duplication:      0/100 ❌  (30.0% fragments cloned, 3 groups)
```

The clamped value reaches every output surface: JSON (`domain/analyze.go:263`), terminal (`cmd/pyscn/analyze.go:549`), HTML (`service/analyze_html.go:435`, `:954`), and CSV (`service/analyze_formatter.go:202`).

## Root cause

`app/analyze_usecase.go:991`:

```go
if totalFragments > 0 && totalClones > 0 {
    summary.CodeDuplication = math.Min(domain.DuplicationThresholdHigh, float64(totalClones)/float64(totalFragments)*100)
}
```

`DuplicationThresholdHigh = 30.0` (`polyscan/core@v0.2.4/domain/scoring.go:20`, the version pinned at `go.mod:24`) is reused for two different jobs: as this input cap, and as the saturation point of the scoring curve. The comment directly above the line states it "Measures what proportion of all code fragments are involved in duplication" — which the clamp makes false.

The clamp is unnecessary because `DuplicationPenalty` (`polyscan/core@v0.2.4/domain/scoring.go:96-108`) already caps itself:

```go
penaltyRange := DuplicationThresholdHigh - DuplicationThresholdLow
penalty := (duplicationPercent - DuplicationThresholdLow) / penaltyRange * 20.0
if penalty > 20.0 {
    penalty = 20.0
}
```

So `DuplicationPenalty(30.0) == DuplicationPenalty(52.46) == 20`. `calculateDuplicationPenalty` (`domain/analyze.go:433`) just forwards to it.

## Repro

Any project above ~30% fragment-clone ratio. Verified with a synthetic case (5 near-identical functions among 15 fragments, true ratio 33.3%) run through the real binary — it also reported exactly `30.0`, confirming the clamp fires generically rather than being specific to one repo.

## Proposed change

Drop the `math.Min` at `app/analyze_usecase.go:991` and let `CodeDuplication` carry the true ratio:

```go
summary.CodeDuplication = float64(totalClones) / float64(totalFragments) * 100
```

`DuplicationPenalty`'s own clamp preserves the existing saturation behaviour, so **no score changes**. Only the reported percentage becomes accurate.

## Verification

Survived refutation — an independent pass confirmed the redundancy claim by reading `DuplicationPenalty`, reproduced the clamp on a synthetic repo with the real binary, audited every consumer of `CodeDuplication` (none assume a `<= 30` bound; the only width-styled HTML bar uses a separately-normalized per-type percentage at `analyze_html.go:993`), and found no doc, comment, or CHANGELOG entry asserting the cap is intentional. `git log -S` shows it was carried over unchanged in commit `4faf98f` ("fix(score): change duplication metric from group density to fragment ratio"), i.e. copy-paste carryover rather than a deliberate decision. `domain/analyze.go:325-326` validates only `0-100`, with no `<= 30` constraint anywhere.

## Impact / effort / confidence

- **Impact**: medium — no score changes, but the headline duplication figure shown in every report format is wrong for exactly the projects with the most duplication, and it silently destroys the ability to track duplication over time or compare projects once past 30%.
- **Effort**: small (one line)
- **Confidence**: high

pyscn version: `v1.30.0-2-g5eb0bd9`

Found via the pyscn-enhancement-audit skill in repo `jodal/pykka@30af2130c66936762cc68ef06ecfe850d8bc9fee`.

7487:
## Summary
`code_duplication_percentage` was clamped to `DuplicationThresholdHigh` (30.0) before being stored in the summary, so any project whose true fragment-clone ratio exceeds 30% reported exactly `30.0` across JSON, terminal, HTML, and CSV output. The clamp is redundant for scoring: `DuplicationPenalty` in polyscan/core already saturates at the same threshold, so `DuplicationPenalty(30.0) == DuplicationPenalty(52.5) == 20`. Removing the pre-storage clamp changes no score — only the reported percentage becomes accurate.

## Type of Change

## Related Issues
Fixes #724

## Changes
- Drop the `math.Min(domain.DuplicationThresholdHigh, ...)` clamp in `calculateSummary` (`app/analyze_usecase.go`) and store the true `totalClones/totalFragments` ratio
- Add a regression test asserting the summary carries an above-30% ratio unclamped while `DuplicationScore` stays at its saturated value (no score change)

## Testing
- [x] `go test ./app/ ./domain/ ./service/ ./cmd/...` passes locally
- [x] `gofmt` / `go vet` clean on touched packages
- [x] Added tests for new functionality (the new test fails against the old clamped code and passes with the fix)
- [ ] Full `make test` / `make lint` not run (only `app` touched; its tests and dependent packages pass)

## Documentation
- [x] Updated the inline comment at the changed line
- [ ] No README/doc changes needed (reporting-only fix, no interface change)

## Additional Context
`TotalClones` counts distinct fragments involved in clones (`len(seenFragments)` in `internal/analyzer/clone_detector.go`), so the ratio is bounded by 100 and the existing `0-100` validation in `domain/analyze.go` still holds. No consumer of `CodeDuplication` assumes a `<= 30` bound; the HTML per-type bars use separately normalized percentages.

Written with the assistance of Claude Code.

### DIFF
--- a/app/analyze_usecase.go
+++ b/app/analyze_usecase.go
@@ -988,7 +988,9 @@ func (uc *AnalyzeUseCase) calculateSummary(summary *domain.AnalyzeSummary, respo
 		totalClones := response.Clone.Statistics.TotalClones
 
 		if totalFragments > 0 && totalClones > 0 {
-			summary.CodeDuplication = math.Min(domain.DuplicationThresholdHigh, float64(totalClones)/float64(totalFragments)*100)
+			// Store the true ratio; DuplicationPenalty saturates at
+			// DuplicationThresholdHigh on its own, so no clamp is needed here.
+			summary.CodeDuplication = float64(totalClones) / float64(totalFragments) * 100
 		}
 	}
 
--- a/app/analyze_usecase_test.go
+++ b/app/analyze_usecase_test.go
@@ -3,6 +3,7 @@ package app
 import (
 	"context"
 	"errors"
+	"math"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -1190,5 +1191,35 @@ func TestBuildComplexityTaskRequest_UsesExecutionConfigBooleans(t *testing.T) {
 	}
 	if req.ShowDetails == nil || !*req.ShowDetails {
 		t.Error("ShowDetails: expected explicit true from execution config")
+	}
+}
+
+func TestCalculateSummaryDuplicationPercentageNotClamped(t *testing.T) {
+	// Regression for #724: CodeDuplication must carry the true fragment-clone
+	// ratio even above the 30% scoring saturation point, instead of being
+	// clamped to DuplicationThresholdHigh before storage.
+	summary := &domain.AnalyzeSummary{}
+	response := &domain.AnalyzeResponse{
+		Clone: &domain.CloneResponse{
+			Statistics: &domain.CloneStatistics{
+				TotalFragments: 61,
+				TotalClones:    32,
+			},
+		},
+	}
+
+	(&AnalyzeUseCase{}).calculateSummary(summary, response)
+
+	want := 32.0 / 61.0 * 100 // 52.46%, above the old 30.0 cap
+	if math.Abs(summary.CodeDuplication-want) > 1e-9 {
+		t.Errorf("CodeDuplication = %f, want %f (unclamped ratio)", summary.CodeDuplication, want)
+	}
+	if summary.Grade == "N/A" {
+		t.Fatal("health score calculation failed for a valid summary")
+	}
+	// Scoring is unchanged: DuplicationPenalty saturates at 30%, so any ratio
+	// at or above it yields the same (zero) duplication score.
+	if summary.DuplicationScore != 0 {
+		t.Errorf("DuplicationScore = %d, want 0 (penalty saturated at 30%%)", summary.DuplicationScore)
 	}
 }
--- a/website/docs/fr/output/health-score.md
+++ b/website/docs/fr/output/health-score.md
@@ -120,7 +120,7 @@ Source : `domain/analyze.go:283-296`. Facteur de normalisation dérivé dans `Ca
 
 ### Duplication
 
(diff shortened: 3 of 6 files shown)


DaisukeYoda (approved):
Hello @7487, thank you for contributing.

Reviewed the change and the fix looks correct and minimal.

One follow-up: `website/docs/output/health-score.md` (en/ja/fr/zh, lines 123 and 145) still documents the clamp (`min(DuplicationThresholdHigh, ...)`) and the capped-at-30 claim. Please update those four pages in this PR or a follow-up so docs and code agree.

LGTM with that addressed.

7487:
Thanks for the review @DaisukeYoda. Updated the four health-score doc pages (en/ja/fr/zh) in 3c0a562: the upstream-computation formula no longer shows the `min(DuplicationThresholdHigh, ...)` clamp, and the input description now notes the value is unclamped with saturation handled by the penalty formula. The scoring/saturation sections were already accurate, so they are unchanged.

DaisukeYoda:
Thanks! LGTM!

Issue #637, opened 2026-07-09
[BUG][auto] cbo-counts-same-file-class-as-dependency: peer class in same file counted as CBO dependent class


DaisukeYoda:
### Bug Description

pyscn's CBO analyzer counts a class defined in the same file as a "dependent class" of another class in that file. In `tfeldmann/organize`, the `Confirm` class (defined at `organize/output/default.py:53`) appears in `Default`'s `DependentClasses` list (defined at `organize/output/default.py:59`). A same-file peer class is not an external dependency — it's an internal implementation detail.

### Repro Steps

Create a minimal Python file with two classes:

```python
# same_file_classes.py
class Helper:
    def do(self) -> str:
        return "done"

class Main:
    def __init__(self, h: Helper) -> None:
        self.h = h
    def execute(self) -> str:
        return self.h.do()
```

Run `pyscn analyze --json --no-open .` and inspect `Main`'s dependent classes:

```bash
jq '.cbo.Classes[] | select(.Name == "Main") | .Metrics.DependentClasses'
```

### Expected Behavior

`DependentClasses` of `Main` should be empty or only list external imports, not `["Helper"]`. A class in the same file does not contribute to external coupling.

### Actual Output

From `tfeldmann/organize` (Default class):
```json
["ChangeDetector", "Confirm", "Console", "Panel", "Path", "Status", "Theme"]
```

`Confirm` is defined at line 53 in the same file — it should not appear as an external dependency.

### Priority Rationale (P2)

Visible but bounded inaccuracy: inflates coupling by 1 for one class in this repo. The pattern reproduces in any multi-class file. Bounded fix: filter out symbols whose definition location is in the same file as the class being analyzed.

### Environment
- **pyscn**: v1.26.1-12-gdc83de5
- **Repo**: tfeldmann/organize@36a54572488d89dc9279d79848ecc067b632f1a5
- Found via FP-audit skill

PR #727, opened 2026-09-01
fix: record dependency kind for imported names in CBO breakdown

### LINKED ISSUE #692: [BUG][auto] cbo-breakdown-collapses-to-import: CBO dependency-kind breakdown counters are always zero because the import check short-circuits
## Description

`CBOMetrics` exposes a breakdown of *how* a class couples to each dependency:
`InheritanceDependencies`, `TypeHintDependencies`, `InstantiationDependencies`,
`AttributeAccessDependencies`, and `ImportDependencies`. In practice the first
four are effectively always `0` and the entire `CouplingCount` is attributed to
`ImportDependencies`, because any dependency reached through an imported name is
bucketed as an import and its actual `kind` is discarded.

Root cause is the early `return` in `addDependency`
(`internal/analyzer/cbo.go:1064-1069`):

```go
dependencies.all[className] = true

if a.isImportedDependency(className) {
    dependencies.imports[className] = true
    return                       // <-- kind is dropped here
}

switch kind {
case dependencyKindInheritance:
    dependencies.inheritance[className] = true
case dependencyKindTypeHint:
    dependencies.typeHints[className] = true
case dependencyKindInstantiation:
    dependencies.instantiations[className] = true
case dependencyKindAttributeAccess:
    dependencies.attributeAccesses[className] = true
case dependencyKindImport:
    dependencies.imports[className] = true
}
```

The import check wins over every other category. Since cross-module coupling in
Python essentially always arrives via an import, and same-file top-level peers
are already dropped as internal (#637), the four non-import counters are
unreachable for realistic code. The breakdown carries no information beyond the
flat `DependentClasses` list.

These fields are user-visible in all three formatters — JSON, the text report
(`service/cbo_formatter.go:187-200`), and the CSV columns
(`service/cbo_formatter.go:253-257`).

## Repro

```python
# base.py
class Base:
    pass


class Helper:
    pass
```

```python
# child.py
from base import Base, Helper


class Child(Base):
    def __init__(self) -> None:
        self.helper = Helper()

    def make(self) -> Helper:
        return Helper()
```

```bash
pyscn analyze --json --no-open --select cbo .
```

## Expected

`Child` inherits `Base` and instantiates `Helper` twice, so roughly:

```
InheritanceDependencies:   1     (Base)
InstantiationDependencies: 1     (Helper)
TypeHintDependencies:      1     (Helper, as the return annotation)
```

## Actual

```json
{
  "Name": "Child",
  "FilePath": "child.py",
  "StartLine": 4,
  "EndLine": 9,
  "Metrics": {
    "CouplingCount": 2,
    "InheritanceDependencies": 0,
    "TypeHintDependencies": 0,
    "InstantiationDependencies": 0,
    "AttributeAccessDependencies": 0,
    "ImportDependencies": 2,
    "DependentClasses": ["Base", "Helper"]
  },
  "RiskLevel": "low",
  "IsAbstract": false,
  "BaseClasses": ["Base"]
}
```

`InheritanceDependencies` is `0` even though `BaseClasses` in the very same
record lists `Base`. The two fields directly contradict each other.

## Real-world confirmation

All 18 classes in `jodal/pykka` show the identical shape — every non-import
counter `0`, `ImportDependencies == CouplingCount`. Two representative records
straight from the analyze JSON:

```json
{
  "Name": "ThreadingActor",
  "FilePath": "src/pykka/_threading.py",
  "StartLine": 122,
  "Metrics": {
    "CouplingCount": 3,
    "InheritanceDependencies": 0,
    "TypeHintDependencies": 0,
    "InstantiationDependencies": 0,
    "AttributeAccessDependencies": 0,
    "ImportDependencies": 3,
    "DependentClasses": ["Actor", "queue.Queue", "threading.Thread"]
  },
  "BaseClasses": ["Actor"]
}
```

`class ThreadingActor(Actor):` is the only reference to `Actor` in that class
body — no type hint, no instantiation, no attribute access. It is an unambiguous
base-class reference, reported as an import dependency.

```json
{
  "Name": "Actor",
  "FilePath": "src/pykka/_actor.py",
  "StartLine": 36,
  "EndLine": 397,
  "Metrics": {
    "CouplingCount": 5,
    "InheritanceDependencies": 0,
    "TypeHintDependencies": 0,
    "InstantiationDependencies": 0,
    "AttributeAccessDependencies": 0,
    "ImportDependencies": 5,
    "DependentClasses": [
      "ActorDeadError", "ActorRef", "ActorRegistry",
      "messages._ActorStop", "threading.Event"
    ]
  },
  "BaseClasses": ["abc.ABC"]
}
```

`ActorRef` is instantiated (`self._actor_ref = ActorRef(self)`,
`src/pykka/_actor.py:218`) yet `InstantiationDependencies` is `0`.

## Note on the existing tests

`internal/analyzer/cbo_test.go` currently asserts these counters are `0` in many
places (lines 517, 529, 536, 543-544, 553, 599, 630, 661, 817-818, 1027, 1571,
1608, 1714), so the always-zero behavior is baked into the suite. Only
`app/cbo_usecase_test.go:114-117` uses non-zero values, and that is a
hand-constructed fixture that never goes through the analyzer. Fixing this needs
those assertions revisited.

One design question to settle first: whether a dependency should be counted in
both its `kind` bucket and the `imports` bucket (buckets then overlap and no
longer sum to `CouplingCount`), or whether `kind` should simply win over the
import check.

## Priority: P2

Per the rubric, a visible but bounded inaccuracy — fields internally inconsistent
while the gist is right. The headline `CouplingCount` and `DependentClasses` are
correct; only the categorization breakdown is wrong. Same shape as #643
(`CouplingBetweenObjects` always null), which was filed P2.

## Found via

The FP-audit skill in repo `jodal/pykka@0a7276126232462d8fb29372c1c418961c77288e`,
pyscn version `v1.29.0` (commit `81b7ec0`).

7487:
## Summary

`addDependency` returned early when the dependency name was imported, so the dependency landed in the `imports` bucket and its actual `kind` was discarded. Since cross-module coupling in Python essentially always arrives via an import (and same-file top-level peers are dropped as internal per #637), `InheritanceDependencies`, `TypeHintDependencies`, `InstantiationDependencies`, and `AttributeAccessDependencies` were effectively always `0`, contradicting `BaseClasses` in the same record.

## Type of Change

## Related Issues
Fixes #692

## Changes
- `internal/analyzer/cbo.go`: drop the early return — an imported dependency is now recorded in both the `imports` bucket (provenance) and its usage-kind bucket. This is the first option from the design question in #692: buckets may overlap and do not sum to `CouplingCount`. The alternative (kind wins over the import check) would leave `ImportDependencies` permanently zero instead, since no call site passes `dependencyKindImport` — so overlapping buckets is the only variant that keeps all five counters meaningful.
- `internal/analyzer/cbo_test.go`: regression test with the repro from the issue (`Child(Base)` inheriting, instantiating, and annotating imported names). Fails on `main`, passes with the fix.

The existing zero assertions in `cbo_test.go` all cover classes with `CouplingCount == 0`, so none needed changes.

On the issue's repro, `pyscn analyze --json --select cbo` now reports for `Child`: `inheritance_dependencies: 1, type_hint_dependencies: 1, instantiation_dependencies: 1, import_dependencies: 2, coupling_count: 2`.

## Testing
- [x] `make test` passes locally (`go test ./...`)
- [x] `make lint` passes locally (`go vet` + gofmt clean; local golangci-lint is v1 and can't read the repo's v2 config)

## Documentation
- [x] Updated godoc comments

### DIFF
--- a/internal/analyzer/cbo.go
+++ b/internal/analyzer/cbo.go
@@ -1062,9 +1062,13 @@ func (a *CBOAnalyzer) addDependency(dependencies *cboDependencies, className str
 
 	dependencies.all[className] = true
 
+	// A dependency reached through an imported name is recorded in the imports
+	// bucket for its provenance AND in the bucket for its usage kind below.
+	// Previously the import check returned early, which made the four kind
+	// counters unreachable for cross-module coupling (see #692). Buckets may
+	// therefore overlap and do not sum to CouplingCount.
 	if a.isImportedDependency(className) {
 		dependencies.imports[className] = true
-		return
 	}
 
 	switch kind {
--- a/internal/analyzer/cbo_test.go
+++ b/internal/analyzer/cbo_test.go
@@ -1781,6 +1781,47 @@ class TOKENS:
 	assert.Contains(t, tokens.DependentClasses, "ast.Attribute")
 }
 
+func TestCBOAnalyzer_ImportedDependenciesKeepKindBreakdown(t *testing.T) {
+	// A dependency reached through an imported name must be counted under its
+	// actual usage kind (inheritance, type hint, instantiation, ...) as well
+	// as in the imports bucket. Before #692 the import check returned early
+	// and the kind was dropped, so the four kind counters were always zero
+	// for cross-module coupling.
+	pythonCode := `
+from base import Base, Helper
+
+class Child(Base):
+    def __init__(self) -> None:
+        self.helper = Helper()
+
+    def make(self) -> Helper:
+        return Helper()
+`
+
+	ast, err := parseCode(pythonCode)
+	require.NoError(t, err)
+
+	results, err := NewCBOAnalyzer(DefaultCBOOptions()).AnalyzeClasses(ast, "test.py")
+	require.NoError(t, err)
+
+	resultMap := make(map[string]*CBOResult)
+	for _, result := range results {
+		resultMap[result.ClassName] = result
+	}
+
+	child := resultMap["Child"]
+	require.NotNil(t, child)
+	assert.Equal(t, 2, child.CouplingCount)
+	assert.ElementsMatch(t, []string{"Base", "Helper"}, child.DependentClasses)
+	assert.Equal(t, 1, child.InheritanceDependencies, "Base is inherited")
+	assert.Equal(t, 1, child.TypeHintDependencies, "Helper is a return annotation")
+	assert.Equal(t, 1, child.InstantiationDependencies, "Helper is instantiated")
+	// Both dependencies arrive via the import, so the imports bucket keeps
+	// recording their provenance. Buckets overlap by design and do not sum
+	// to CouplingCount.
+	assert.Equal(t, 2, child.ImportDependencies)
+}
+
 func parseCode(code string) (*parser.Node, error) {
 	p := parser.New()
 	ctx := context.Background()


DaisukeYoda (approved):
Thanks! LGTM!

PR #729, opened 2026-09-01
fix: stop counting elif arms as extra nesting levels

### LINKED ISSUE #728: [BUG][auto] elif-chain-nesting-inflation: each elif arm counted as a nesting level, flat dispatch reported nesting_depth 9 and risk high
## Bug

`nesting_depth` counts every `elif` arm of a flat `if/elif/elif/...` chain as one more nesting level. A function whose deepest real nesting is 2 is reported with `nesting_depth: 9`, and because `AssessRiskLevel` promotes any function with `nesting_depth > NestingDepthThreshold` to `high`, a plain dispatch table gets a **high** risk grade that its cyclomatic complexity (13, below the medium threshold of 19) does not justify.

## Root cause

`internal/parser/ast_builder.go` links each `elif` to the next through `Orelse` (`elifNodes[i].Orelse = []*Node{elifNodes[i+1]}`), so a chain of N arms becomes N nested `If`/`ElifClause` nodes.

`traverseForNesting` in `internal/analyzer/nesting_depth.go` then recurses into `node.Orelse` at `newDepth` (the parent's depth + 1), and `isNestingNode` returns true for `NodeIf` and `NodeElifClause`, so each arm adds one more level. `NodeElseClause` is already treated as "same level"; `elif` should be too.

## Repro

```python
def dispatch(kind):
    if kind == "a":
        return 1
    elif kind == "b":
        return 2
    elif kind == "c":
        return 3
    elif kind == "d":
        return 4
    elif kind == "e":
        return 5
    elif kind == "f":
        if kind:          # the only real nesting: depth 2
            return 6
    raise NotImplementedError
```

```
```

## Expected

`nesting_depth: 2` (the inner `if` inside the last arm). Risk level driven by cyclomatic complexity alone.

## Actual output

From `aiomisc_log/__init__.py` in aiomisc, `create_logging_handler` is a 9-arm `elif` chain over a `LogFormat` enum; its deepest real nesting is 2 (`elif ... syslog:` -> `if os.path.exists("/dev/log"):`):

```json
{
  "name": "create_logging_handler",
  "start_line": 27,
  "end_line": 79,
  "risk_level": "high",
  "metrics": {
    "complexity": 13,
    "cognitive_complexity": 18,
    "nesting_depth": 9,
    "if_statements": 12,
    "loop_statements": 0,
    "exception_handlers": 0,
    "sloc": 47
  }
}
```

`nesting_depth` equals the number of `elif` arms plus the one real inner `if`.

## Environment

- pyscn `v1.30.0-2-g5eb0bd9`
- Found via the FP-audit skill in repo `aiokitchen/aiomisc@22d3bc91f27e1bc0b839f3c907dc07b4022d00bb`
- Same pattern previously seen in `apiflask/apiflask` (9 arms reported as depth 10) and `tlsfuzzer/tlslite-ng` (28-arm chain reported as depth 29)

## Priority

P2: bounded metric inaccuracy, but it flips `risk_level` to `high` for ordinary dispatch functions. Fix is local to `traverseForNesting` (pass `currentDepth` instead of `newDepth` when the `Orelse` child is an `elif`), so it is a reasonable good first issue.

7487:
## Summary
`nesting_depth` counted every `elif` arm of a flat `if/elif/...` chain as one more nesting level, because the AST builder chains elif arms through `Orelse` and `isNestingNode` treated `NodeElifClause` as a nesting construct. A flat dispatch chain with one real inner `if` was reported with `nesting_depth` = arm count + 1, which promoted `risk_level` to `high` via the nesting threshold.

This treats `NodeElifClause` like `NodeElseClause` — same level as the parent `if` — so only real nesting counts.

## Type of Change

## Related Issues
Fixes #728

## Changes
- `internal/analyzer/nesting_depth.go`: move `NodeElifClause` from the nesting case to the same-level case in `isNestingNode`
- `internal/analyzer/nesting_depth_test.go`: regression test parsing the issue's repro (flat 5-arm elif chain with one nested `if` now reports depth 2), plus an `ElifClause` entry in `TestIsNestingNode`
- `internal/analyzer/complexity_test.go`: `TestCalculateComplexity_ClassSuiteOwnsControlFlow` expected depth 2 for a flat class-body `if/elif/else` — that expectation encoded the bug; updated to 1

## Testing
- [x] `go test ./...` passes locally
- [x] `gofmt` / `go vet` clean on the touched package (local golangci-lint is v1 and can't read the repo's v2 config)
- [x] Manual testing completed: built the binary and ran `analyze --json` on the issue repro — `dispatch` now reports `nesting_depth: 2`, `risk_level: low` (was 7/high)

### DIFF
--- a/internal/analyzer/complexity_test.go
+++ b/internal/analyzer/complexity_test.go
@@ -365,8 +365,9 @@ class Config:
 	if classResult.IfStatements != 2 || classResult.LoopStatements != 1 {
 		t.Fatalf("class statement metrics = if:%d loop:%d, want 2/1", classResult.IfStatements, classResult.LoopStatements)
 	}
-	if classResult.NestingDepth != 2 {
-		t.Fatalf("class nesting depth = %d, want 2", classResult.NestingDepth)
+	// The if/elif/else chain is flat, so the deepest nesting is 1 (issue #728).
+	if classResult.NestingDepth != 1 {
+		t.Fatalf("class nesting depth = %d, want 1", classResult.NestingDepth)
 	}
 }
 
--- a/internal/analyzer/nesting_depth.go
+++ b/internal/analyzer/nesting_depth.go
@@ -115,12 +115,14 @@ func isNestingNode(node *parser.Node) bool {
 	}
 
 	switch node.Type {
-	case parser.NodeIf, parser.NodeFor, parser.NodeAsyncFor, parser.NodeWhile, parser.NodeWith, parser.NodeAsyncWith, parser.NodeTry, parser.NodeExceptHandler, parser.NodeMatch, parser.NodeMatchCase, parser.NodeElifClause,
+	case parser.NodeIf, parser.NodeFor, parser.NodeAsyncFor, parser.NodeWhile, parser.NodeWith, parser.NodeAsyncWith, parser.NodeTry, parser.NodeExceptHandler, parser.NodeMatch, parser.NodeMatchCase,
 		parser.NodeLambda, parser.NodeListComp, parser.NodeSetComp, parser.NodeDictComp, parser.NodeGeneratorExp:
 		// These nodes increase nesting depth
 		return true
-	case parser.NodeElseClause:
-		// else doesn't add nesting, it's at same level
+	case parser.NodeElifClause, parser.NodeElseClause:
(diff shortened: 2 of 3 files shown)


DaisukeYoda (approved):
LGTM!

PR #735, opened 2026-09-04
fix(analyzer): cap semantic Type-4 similarity below the Type-1 threshold

### LINKED ISSUE #733: [BUG][auto] clone-type4-similarity-bypasses-nontextual-cap: Type-4 similarity is uncapped while Type-1/2/3 are clamped, inverting clone ranking
## Description

Non-textual clone pairs are deliberately capped just below the Type-1 threshold so they "never report a Type-1-level similarity". Type-4 semantic pairs **bypass that cap**, so they are the only clone type that can report a high similarity — and they do, routinely.

The result is that `similarity` is not comparable across clone types. Sorting clone pairs by similarity systematically ranks the weakest evidence (CFG shape matches) above the strongest (normalized-AST matches).

## Repro

Any repo where semantic analysis produces Type-4 pairs. From a single run:

```bash
jq '[.clone.clone_pairs[] | select(.type==2) | .similarity] | unique' analyze.json
jq '[.clone.clone_pairs[] | select(.type==4) | .similarity] | max' analyze.json
```

## Expected

`similarity` means the same thing regardless of `type`, so values are comparable across types — or, if the cap is intentional, it applies uniformly.

## Actual

In one run over `dajiaji/pyseto@628d6a3` (13 Type-2 pairs, 332 Type-4 pairs):

```json
// distinct Type-2 similarities
[0.7566371681415929, 0.79875, 0.8499999999999999]

// max Type-4 similarity
0.9777304534133096

// Type-4 pairs at or above the Type-1 threshold (0.85)
90
```

11 of 13 Type-2 pairs sit at exactly `0.8499999999999999` = `math.Nextafter(0.85, 0)` — the cap constant, not a measurement. Meanwhile 90 Type-4 pairs exceed 0.85 freely, topping out at 0.978.

Concretely, in that run an **unrelated** pair (`key_nist.py:214-239` PBKDF2 key-unwrap vs `key_nist.py:329-347` ECDH seal — no shared logic) scored **0.978**, while a genuine 4-way-copied twin (`v3.py:270-282` vs `v4.py:114-129`, the same `to_peer_paserk_id` body) scored **0.974**. The false pair outranks the real one.

## Root cause

`compareFragmentsWithClassifier` (`internal/analyzer/clone_detector.go:1097`) returns early for semantic Type-4 and passes `result.Similarity` straight through:

```go
if result.CloneType == Type4Clone && result.Analyzer == "semantic" {
    return &ClonePair{
        ...
        Similarity: result.Similarity,   // uncapped
        ...
    }
}

cloneType, similarity := cd.classifyClonePair(fragment1, fragment2, similarity)  // capped
```

Every other path goes through `classifyClonePair`, which applies `capNonTextualSimilarity` (`polyscan/core@v0.2.4/clone/classifier.go:88-101`):

```go
// capNonTextualSimilarity caps structural similarity just below the Type-1
// threshold so that pairs without an exact textual match never report a
// Type-1-level similarity.
```

with `DefaultType1CloneThreshold = 0.85` (`domain/defaults.go:15`).

So the cap's own stated invariant — non-textual pairs never report Type-1-level similarity — is violated by exactly the type with the weakest evidence.

## Impact

- Any "top clones" listing or `sort_by: similarity` ordering is misleading: coincidental CFG matches float to the top.
- Type-2 similarity is nearly useless as a signal, since most values collapse onto one constant and cannot distinguish a near-identical pair from a borderline one.
- Cross-repo: the same `0.8499999999999999` clamp on all Type-2 pairs was observed in earlier audits of `life4/deal` and `nosarthur/gita`.

Either direction of fix resolves it (cap Type-4 too, or stop capping and report measured similarity everywhere with type carrying the confidence) — but the two paths should agree.

## Priority: P2

Visible but bounded inaccuracy (per rubric): the clone pairs themselves are still reported and the gist is right, but one output field is internally inconsistent between code paths in a way that inverts ranking. Not P1, since no nonexistent entity is cited and no data is lost.

## Environment

pyscn `v1.30.1` (commit `2c1ff91`)

## Found via

FP-audit skill in repo `dajiaji/pyseto@628d6a34e0ede88abcb8234108272eda6745468d`

7487:
## Summary
`compareFragmentsWithClassifier` returns early for semantic Type-4 pairs and passes `result.Similarity` straight through, while every other path goes through `classifyClonePair` -> `capNonTextualSimilarity`, which clamps non-textual similarity just below the Type-1 threshold. Type-4 was therefore the only clone type that could report Type-1-level similarity, so sorting clone pairs by similarity ranked coincidental CFG-shape matches above genuine normalized-AST matches (see #733 for a concrete inversion on `dajiaji/pyseto`).

This applies the same cap on the semantic Type-4 early return, restoring the cap's stated invariant: pairs without an exact textual match never report a Type-1-level similarity. A semantic Type-4 result can never be an exact textual match, since the classifier checks Type-1 first.

## Type of Change

## Related Issues
Fixes #733

## Changes
- `internal/analyzer/clone_detector.go`: add `CloneDetector.capNonTextualSimilarity` (mirrors the core pair classifier's cap using the detector's configured thresholds) and apply it to the semantic Type-4 early return in `compareFragmentsWithClassifier`.
- `internal/analyzer/clone_detector_test.go`: add `TestCloneDetector_SemanticType4SimilarityIsCapped` covering both the capped (raw 0.98 -> just below Type-1 threshold) and pass-through (0.7 unchanged) cases.

## Testing
- [x] `go build ./...`, `go vet ./internal/analyzer/`, `go test ./internal/analyzer/ -count=1` pass locally
- [ ] `make lint` (local golangci-lint is v1, config needs v2; `gofmt -l` clean on touched files)

## Documentation
- [x] Updated godoc comments

## Additional Context
The issue notes either direction (cap Type-4 too, or uncap everything) resolves the inconsistency. This PR takes the capping direction since the invariant is already documented on `capNonTextualSimilarity` in polyscan core and applied to Type-1/2/3; uncapping would change reported similarities for all existing clone types.


DaisukeYoda (changes requested):
@7487 

`capNonTextualSimilarity` duplicates the implementation in the `polyscan/core` `PairClassifier`, creating two places where threshold behavior and future changes must be maintained. Please keep responsibility for the cap in core: expose an API such as `PairClassifier.CapNonTextualSimilarity`, then call it here:

```go
Similarity: cd.pairClassifier.CapNonTextualSimilarity(result.Similarity),
```

This lets us remove the helper introduced in this PR. The separate issue of the cap discarding raw semantic-score ordering can be tracked independently; longer term, display `Similarity` and the internal ranking score should be separated.

7487:
Agreed, the cap should have one owner. The helper in `polyscan/core` is currently unexported (`capNonTextualSimilarity`), so I opened ludo-technologies/polyscan#110 to export it as `PairClassifier.CapNonTextualSimilarity` (rename only, no behavior change; `ClassifyPair` still applies it internally).

Once that lands in a tagged `core` release, I'll push a follow-up here that bumps `go.mod`, replaces the local helper with `cd.pairClassifier.CapNonTextualSimilarity(result.Similarity)`, and keeps the test. The existing `TestCloneDetector_SemanticType4SimilarityIsCapped` covers the swap unchanged since it asserts on the reported similarity, not the helper.

DaisukeYoda:
Merged polyscan#110 and tagged core/v0.2.7. Once you bump go.mod and swap the local helper for the exported method this is good to go; the test already asserts on the reported similarity, so no changes needed there.

7487:
Pushed two commits:

- 685a838 bumps `polyscan/core` to v0.2.7 and swaps the local helper for `cd.pairClassifier.CapNonTextualSimilarity(result.Similarity)`. `TestCloneDetector_SemanticType4SimilarityIsCapped` is unchanged.
- bb140de: the bump also pulls in core v0.2.6 (polyscan#105), and pyscn aliases `DuplicationThresholdHigh` from core, so `DuplicationPenalty` now saturates at 60% instead of 30%. That failed 7 cases of `TestAnalyzeSummary_CalculateHealthScore` and `TestCalculateSummaryDuplicationPercentageNotClamped`. This commit updates those expectations, the stale 30% comments and the health-score docs table, mirroring polyscan#106. pyscn's default Type-2/3 thresholds live in `domain/defaults.go`, so they are not affected. If you would rather land the v0.2.6 adoption in its own PR, say so and I will drop this commit here.

`go build ./... && go vet ./... && go test ./... -count=1` pass locally.

DaisukeYoda (approved):
Verified locally and reviewed the recalculated expectations, all consistent. Keeping the v0.2.6 adoption here is fine since it is inseparable from the bump; I'll call out the duplication scale change in the release notes. Thanks for the careful follow-through across both repos.

PR #740, opened 2026-09-05
fix: surface project root import diagnostics

### LINKED ISSUE #738: Wrong project-root inference silently degrades import resolution and inflates the score
Found via a user's CI workaround in DOI-USGS/dataretrieval-python (`.github/workflows/code-health.yml`, see DOI-USGS/dataretrieval-python#415).

When pyscn infers the wrong project root, only part of the imports resolve. Since most of the score is dependency-derived, a degraded run reports a *higher* score than a correct one. Nothing in the output signals this, so the user added a sanity step that compares `Summary.ProjectRoot` to the checkout and counts `TotalDependencies`.

Expected:
- Emit a visible warning when the inferred root differs from the analyzed directory / cwd, or when a large share of internal imports fail to resolve.
- Expose resolved vs. unresolved import counts in the summary so the score can be sanity-checked.
- Provide an explicit `--project-root` flag (or config key) so CI does not depend on inference.

mikemikimike:
## Summary / Problem

When pyscn analyzes a subdirectory, marker-based project-root inference can assign modules the wrong qualified names. That silently drops internal dependency edges and can make dependency-derived scores look healthier than the analyzed project really is.

## Changes

- Prefer the discovered `.pyscn.toml` or `pyproject.toml [tool.pyscn]` directory as the project root.
- Add an optional top-level `project_root` configuration key, resolved relative to the configuration file, for projects whose import root differs from the configuration directory.
- Propagate the resolved root through `analyze`, system analysis, and community analysis; remove the `--project-root` CLI flag.
- Compare root diagnostics with the current working directory and direct users to the configuration key when an explicit root is needed.
- Track resolved and unresolved internal import statements in dependency analysis results and summaries.
- Include the import-resolution counts in text, CSV, HTML, JSON, and YAML reports.
- Add regression coverage for configuration discovery, root propagation, diagnostics, resolution counts, warnings, and CLI output; document the configuration behavior.

## Tests

- `go mod verify`
- `go test ./...`
- `go test -race ./service ./cmd/pyscn`
- `go vet ./...`
- `go build ./cmd/pyscn`

All listed checks passed in a Linux Go 1.26.8 container. The repository CI matrix uses Go 1.22 and 1.23 for tests, Go 1.23 for the build, and Go 1.25 with golangci-lint; the golangci-lint action and those exact toolchain versions were not available locally.

## Compatibility / Known limitations

The new configuration key is optional. Without a discovered pyscn configuration file, automatic marker-based project-root inference remains the fallback. `project_root` paths are resolved relative to the configuration file, and the new report fields are additive. Internal import diagnostics exclude standard-library and third-party imports that are outside the analyzed graph.

## Issue link

Fixes #738

https://github.com/ludo-technologies/pyscn/issues/738

### DIFF
--- a/app/analyze_usecase.go
+++ b/app/analyze_usecase.go
@@ -46,8 +46,9 @@ type AnalyzeUseCaseConfig struct {
 	// Clone detection options
 	EnableDFA bool // Enable Data Flow Analysis for enhanced Type-4 detection
 
-	ConfigFile string
-	Verbose    bool
+	ConfigFile  string
+	ProjectRoot string
+	Verbose     bool
 }
 
 // AnalyzeRequestOverrides contains request-scoped values that take precedence
@@ -497,9 +498,13 @@ func (uc *AnalyzeUseCase) executeProject(ctx context.Context, useCaseCfg Analyze
 	// timings recorded by previous runs on this project (if any)
 	estimatedSeconds := uc.estimateTaskSeconds(len(allFiles), useCaseCfg, executionCfg)
 
+	projectRoot := useCaseCfg.ProjectRoot
+	if projectRoot == "" {
+		projectRoot = service.FindProjectRoot(paths)
+	}
 	snapshot := service.BuildAnalysisProjectSnapshot(ctx, analysisFiles, moduleFiles, service.ProjectSnapshotOptions{
 		IncludeRawMetrics: uc.complexityUseCase != nil && !useCaseCfg.SkipComplexity,
-		ProjectRoot:       service.FindProjectRoot(paths),
+		ProjectRoot:       projectRoot,
 	})
 
 	var moduleGraph *service.ProjectModuleGraph
@@ -751,7 +756,8 @@ func (uc *AnalyzeUseCase) createAnalysisTasks(config AnalyzeUseCaseConfig, sourc
 					return nil, fmt.Errorf("prepare system analysis graph: %w", err)
 				}
 				request := domain.SystemAnalysisRequest{
-					Paths:                files,
+					Paths:                append([]string(nil), sourcePaths...),
+					ProjectRoot:          config.ProjectRoot,
 					Recursive:            domain.BoolPtr(executionCfg.Recursive),
 					IncludePatterns:      []string{},
 					ExcludePatterns:      []string{},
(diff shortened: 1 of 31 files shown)


DaisukeYoda (changes requested):
Thanks for taking this on. The resolved/unresolved counts work well: I reproduced a wrong-root case (a `requirements.txt` inside the package) and the ratio warning fired correctly while the score stayed the same, which is exactly what #738 asked for. Two changes before this can go in:

**1. The root warning compares against the wrong thing.**
`systemAnalysisWarnings` compares the inferred root to the analyzed directory, so `pyscn analyze src/` (and the `pyscn analyze dataretrieval` command from the CI that motivated the issue) warns on every run, even when inference is correct and every import resolves. The check in the issue, and the workaround in that CI, compare against cwd. Please compare against cwd instead. That also lets you drop `analysisScopeRoot`, which duplicates the common-parent logic in `FindProjectRoot`.

**2. Prefer a config key over a CLI flag.**
Config discovery already walks up from the analyzed path and stops at the repository boundary, so the directory holding `.pyscn.toml` / `pyproject.toml [tool.pyscn]` is a much stronger root signal than the marker list in `FindProjectRoot`. I'd rather:
- use the config file's directory as the project root whenever a config file is found, and
- add an optional `project_root` key (relative to the config file) for the remaining cases.

That covers every subcommand, not only `analyze`, and gives CI a repeatable setup without per-command flags. Please drop `--project-root` in favor of this.

The rest (counts in all report formats, tests, docs) looks good.

mikemikimike:
Addressed review 5123721374 in `e9e2d989`:

- Root diagnostics now compare the inferred root with the current working directory.
- Removed `--project-root`; configuration discovery uses `.pyscn.toml` or `pyproject.toml [tool.pyscn]`, with optional top-level `project_root` resolved relative to that config file.
- Propagated the resolved root through analyze, system analysis, and community analysis.
- Updated diagnostics, tests, and documentation.

Validation passed locally in a Linux Go 1.26.8 container: `go mod verify`, `go test ./...`, `go test -race ./service ./cmd/pyscn`, `go vet ./...`, and `go build ./cmd/pyscn`.

The new GitHub CI run is currently marked `action_required` with no jobs reported.

DaisukeYoda (approved):
LGTM!

PR #747, opened 2026-09-07
fix: serialize empty report collections as []

### LINKED ISSUE #742: JSON report emits null instead of an empty list for collections with no findings
When a section has zero findings, `analyze --json` writes `null` for the collection (e.g. `complexity.Functions`, `dead_code.files`, `dead_code.files[].functions[].findings`) instead of `[]`. Every consumer has to add `or []` guards; see `extract_findings` in daiki-beppu/youtube-automation `.github/scripts/pyscn-diff-gate.py`.

Expected: empty collections serialize as `[]`. Likely just a matter of initializing the Go slices instead of leaving them nil before marshaling.

7487:
## Summary
`analyze --json` wrote `null` for `complexity.functions`, `dead_code.files`, `dead_code.files[].functions` and `dead_code.files[].functions[].findings` whenever a section had nothing to report. The slices were built with `var x []T` / `append([]T(nil), ...)`, which stay nil when no element is appended, and `encoding/json` marshals a nil slice as `null`. This allocates empty slices at those construction sites so the collections serialize as `[]`.

Repro before the fix (two trivial `.py` files, `pyscn analyze --json --min-complexity 5 .`):

```
complexity.functions = null
dead_code.files      = null
```

After:

```
complexity.functions = []
dead_code.files      = []
```

## Type of Change
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues
Fixes #742

## Changes
- `domain/complexity.go`: `sortComplexityScopes` copies into `make([]FunctionComplexity, 0, n)` instead of `append([]T(nil), ...)`, so a fully filtered `Functions` list is `[]` rather than nil.
- `service/dead_code_service.go`: `filterFiles`, `filterFindingsBySeverity` and `convertToFunctionDeadCode` start from `make(...)` instead of `var`. `filterFiles` also covers `files[].functions` for a file that only has class-scope findings, which previously came out as `null` too.
- Tests: `Analyze` on a trivial file with `MinComplexity: 5` / a clean file asserts the response slices are non-nil and that the JSON contains `"functions":[]` / `"files":[]`; `filterFindingsBySeverity` with no matches returns an empty slice; class-only file keeps a non-nil `Functions`.

`warnings`/`errors` on both responses still serialize as `null` when empty. Left them out since the issue is scoped to finding collections; happy to include them here if you'd prefer.

## Testing
- [x] `go test ./...` passes locally
- [x] `go vet ./...` and `gofmt -l` clean (local golangci-lint is v1 and rejects the v2 config, so CI is the lint check)
- [x] Manual testing completed (repro above)

## Documentation
- [ ] Updated godoc comments


DaisukeYoda (approved):
Looks good. Thanks

PR #750, opened 2026-09-07
feat: add schema_version and document the v1.30.0 key rename

### LINKED ISSUE #744: Document the v1.30.0 JSON key rename (PascalCase to snake_case) as a breaking change
v1.30.0 (#718) changed every key in the analyze JSON report from PascalCase to snake_case (`complexity.Functions[].Metrics` -> `complexity.functions[].metrics`, `system.Summary.ProjectRoot` -> `system.summary.project_root`, and so on). The release notes mention it in one line and nothing flags it as breaking.

Known consumers that break on upgrade:
- lesleslie/crackerjack `crackerjack/parsers/json_parsers.py` (reads `Functions` / `Metrics`)
- daiki-beppu/youtube-automation `.github/scripts/pyscn-diff-gate.py` (reads `Functions`, `RiskLevel`, `FilePath`, `Name`; dependency is `>=1.29,<2`)
- DOI-USGS/dataretrieval-python `.github/workflows/code-health.yml` (reads `system.Summary.ProjectRoot`)

Expected:
- A migration note in CHANGELOG / release notes for v1.30.0 listing the old and new key names.
- Add a top-level `schema_version` to the JSON report so future changes can be detected by consumers.
- Also document that macOS Intel has no wheel since v1.5.1 and point to `brew install pyscn` / `go install` (README install section).

7487:
## Summary
v1.30.0 (#718) renamed every `analyze` JSON/YAML key from Go PascalCase / lowerCamelCase to snake_case, and the only notice was a one-line CHANGELOG bullet plus an HTML comment (invisible when rendered) in the schema docs. This PR documents the rename as a migration, adds a top-level `schema_version` so consumers can detect future key changes, and documents that macOS Intel has no PyPI wheel.

## Type of Change
- [x] Documentation update

## Related Issues
Fixes #744

## Changes
- `domain/analyze.go`: add `AnalyzeSchemaVersion = 1` and a top-level `schema_version` field on `AnalyzeResponse` (JSON and YAML). `app.AnalyzeUseCase.buildResponse` stamps it alongside `version` (moved there from `cmd/pyscn/analyze.go`), so the CLI `--json`/`--yaml` reports and the MCP `analyze` tool in `full` mode all carry both. MCP `full` output previously serialized an empty `version`.
- `CHANGELOG.md`: expand the v1.30.0 **BREAKING** bullet with a migration note (rename rule, commonly read old/new keys, `request`/`config` objects, what did not change).
- `website/docs/output/schemas.md`: turn the hidden HTML comment into a visible "Migrating from 1.29.x" section with an old/new key table, document `schema_version` in the top-level structure, and state that breaking changes increment it.
- `README.md`: note in the install section that PyPI wheels are Apple Silicon only (Intel dropped in v1.5.1, #259) and point Intel Macs to `brew install pyscn` / `go install`.
- `website/docs/getting-started/installation.md`, `website/docs/integrations/python-packaging.md` and their `ja`/`fr`/`zh` copies: the platform tables claimed macOS x86_64 (and Linux/Windows arm64) wheels; corrected to the wheels the release workflow actually builds (`manylinux_2_17_x86_64`, `macosx_11_0_arm64`, `win_amd64`), matching what is on PyPI for 1.31.0.

`schema_version` is an integer starting at `1` for the current snake_case layout and only bumps on a breaking key change; additive fields (which the stability contract already allows) do not bump it. Reports from 1.30.0/1.31.0 have no `schema_version`, so the docs tell consumers to fall back to `version` to tell 1.29.x from 1.30.0+.

Not done here: the GitHub release notes for v1.30.0 (maintainer-editable only), and the `ja`/`fr`/`zh` copies of `schemas.md` — translating the migration section is a bigger job than the platform-table correction, and those pages already lag the English one (they are missing the `diagnostics` and `failures` rows). Happy to fold that in here or follow up separately.

## Testing
- [x] `go test ./domain/... ./app/... ./service/... ./cmd/... ./mcp/...` passes locally
- [x] `gofmt -l` clean, `go vet` clean (local golangci-lint is v1 and cannot read the v2 config; relying on CI for lint)
- [x] Added tests: `TestAnalyzeUseCaseBuildResponseStampsMetadata` (app) and a `"schema_version": 1` assertion in `TestAnalyzeFormatter_Write_JSON` (service)
- [x] Manual: `pyscn analyze --json --output - --select complexity <dir>` emits `"schema_version": 1`; `--yaml` emits `schema_version: 1`

## Documentation
- [x] Updated godoc comments
- [x] Updated README or other docs (if needed)


DaisukeYoda:
@7487 The docs work here is accurate and worth having. I checked the corrected wheel table against the release workflow and it matches what is actually published, and the rename table matches the current struct tags.

Two things I would change before merging. The new schema version is stamped into the response inside the use case while the existing version field is still stamped up in the CLI command, so the MCP full output now carries a schema version but an empty version string, which is awkward given the migration note tells consumers to fall back on version when the schema version is absent. And since the schema version is a compile-time constant, I would rather it not be mutable response state at all, or failing that have both metadata fields stamped in the same place; as it stands any response built outside that one constructor serializes a zero value the spec does not define.

The second is the schema doc's internal consistency: the stability section still states that breaking changes are restricted to major version bumps, immediately above a new migration section documenting a breaking change that shipped in a minor release. That contradiction is essentially why #744 was filed, so it should be reconciled here rather than left standing.

Separately, the platform table correction is a factual fix rather than a translation task, so leaving the ja/fr/zh copies claiming Intel macOS and arm64 Linux/Windows wheels keeps three of the four languages wrong. Worth folding in since it is the same one-line change.

7487:
Thanks — all three are fair. Pushed 04a8d58.

**1. Metadata stamping.** Went with the "same place" option: `response.Version = version.Version` moved out of `generateOutput` in `cmd/pyscn/analyze.go` into `AnalyzeUseCase.buildResponse`, next to `SchemaVersion`. `generateOutput` had the only assignment and one caller, so nothing else had to change, and `app` already depends on `service`, which imports `internal/version`.

This also fixes the empty `version` on its own: MCP `full` mode returns `result` straight from `analyzeUC.ExecuteWithOverrides`, so before this PR it always serialized `"version": ""` — the schema version just made it visible. Now both fields come out of the constructor together.

I did look at making it non-mutable via a `MarshalJSON`/`MarshalYAML`-only type, but `domain.AnalyzeResponse` is round-tripped through `json.Unmarshal` in `service/analyze_formatter_test.go`, so it would need matching unmarshalers to stay decodable — more machinery than the problem is worth. Say the word if you'd rather have it.

Test renamed to `TestAnalyzeUseCaseBuildResponseStampsMetadata` and now asserts both fields. `go test ./app/... ./service/... ./mcp/... ./cmd/... ./domain/...` passes; `pyscn analyze --json` still emits the real version (`{"schema_version": 1, "version": "1.31.0"}` from a local ldflags build).

**2. Stability section.** Reworded so it no longer claims major-only:

> Consumers MUST ignore unknown fields. Breaking changes to the `pyscn analyze` report increment its top-level `schema_version` and are listed as **BREAKING** in the CHANGELOG. They are not restricted to major version bumps — the snake_case key rename described below shipped in the 1.30.0 minor release — so pin against `schema_version` rather than the semantic version.

The page intro said the same thing ("stable across patch releases within the same major version") and now keys off `schema_version` too. I deliberately described current practice rather than writing a new policy — if you'd rather commit to "no breaking changes outside major bumps from now on, 1.30.0 excepted", that's your call to make and I'll reword.

**3. Platform tables.** Folded into `ja`/`fr`/`zh` for both `getting-started/installation.md` and `integrations/python-packaging.md`.

I left the `ja`/`fr`/`zh` copies of `output/schemas.md` alone. Unlike the platform table, the migration section is real translation work, and those three pages already trail the English one (no `diagnostics` or `failures` rows in their top-level field table). Fixing only the stability sentence there would leave them pointing at a `schema_version` they don't document. Happy to translate the whole section in this PR if you want it here rather than in a docs sync.

DaisukeYoda (approved):
Thanks!

PR #751, opened 2026-09-07
fix: group from-import namespace members in CBO

### LINKED ISSUE #732: [BUG][auto] cbo-namespace-member-not-grouped-for-from-import: GroupNamespaceImports misses `from pkg import mod` bindings, inflating CBO
## Description

`GroupNamespaceImports` (default `true`) is documented to stop a single namespace import from adding one CBO edge per member reference:

```go
// canonicalDependencyName collapses references through a namespace alias to the
// alias itself when GroupNamespaceImports is enabled. This prevents a single
// "import libcst as cst" from inflating CBO by one edge per cst.Member usage.
```

It works for `import mod as alias`, but **not** for `from pkg import mod` — which binds a namespace exactly the same way. Every `mod.Member` reference then becomes its own `DependentClasses` entry.

This is the sibling case that the fixes for #517 / #667 did not cover: those addressed the aliased-import form, which now behaves correctly. The `from`-import form still inflates.

## Repro

Two files with **identical class bodies**, differing only in how the namespace is imported:

```python
# a_from_import.py
from pkgmod import serialization

class FromImportUser:
    def run(self):
        return serialization.Encoding.DER, serialization.PrivateFormat.X, serialization.NoEncryption()
```

```python
# b_alias_import.py
import pkgmod.serialization as serialization

class AliasImportUser:
    def run(self):
        return serialization.Encoding.DER, serialization.PrivateFormat.X, serialization.NoEncryption()
```

```bash
```

## Expected

Both classes couple to one namespace (`serialization`), so both should report `coupling_count: 1`.

## Actual

```json
{
  "name": "FromImportUser",
  "coupling_count": 3,
  "dependent_classes": ["serialization.Encoding", "serialization.NoEncryption", "serialization.PrivateFormat"]
}
{
  "name": "AliasImportUser",
  "coupling_count": 1,
  "dependent_classes": ["serialization"]
}
```

Same code, 3x the coupling, purely because of import spelling.

## Root cause

`internal/analyzer/collectImports` (`internal/analyzer/cbo.go:655`) populates `namespaceAliases` only in the `parser.NodeImport` branch, and only when `alias != module`:

```go
case parser.NodeImport:
    ...
    if alias != module {
        imports.namespaceAliases[module] = alias
    }
case parser.NodeImportFrom:
    // ... never writes to imports.namespaceAliases
```

`canonicalDependencyName` (`cbo.go:1091`) collapses a dotted name only when the leading segment resolves through `namespaceAliases`, so `from pkg import mod` bindings never collapse.

The ambiguity worth discussing in the fix: `from pkg import x` may bind either a module or a class, and the two are not distinguishable from the import statement alone — so registering every such binding as a namespace would under-count genuine class dependencies. A usage-shape heuristic (collapse only when the binding is referenced as `x.Attr(...)` with a CapWords `Attr`, never instantiated directly) is probably needed.

## Real-world impact

Found auditing `dajiaji/pyseto@628d6a3`. `V1Public` (`pyseto/versions/v1.py:104-283`) is reported at `coupling_count: 12`, **risk `high`**:

```json
"dependent_classes": [
  "KeyInterface", "NISTKey", "SignError", "VerifyError",
  "hashes.Hash", "hashes.SHA384",
  "padding.MGF1", "padding.PSS",
  "serialization.Encoding", "serialization.NoEncryption",
  "serialization.PrivateFormat", "serialization.PublicFormat"
]
```

8 of the 12 edges are members of just three namespaces, all bound with `from`-imports:

```python
from cryptography.hazmat.primitives import hashes, serialization
from cryptography.hazmat.primitives.asymmetric import padding
```

Collapsed, `V1Public` couples to ~7 things — medium risk, not high. The same pattern puts 4 of this repo's 5 "high coupling" classes over the threshold and drives its Coupling score to 0/100.

## Priority: P2

Visible but bounded inaccuracy (per rubric): the coupling numbers are inflated rather than nonsensical, but the inflation is large enough to move classes across risk tiers and trigger false "reduce coupling" suggestions. `from pkg import module` is the dominant idiom for `cryptography`, `os.path`, `xml.etree`, and most stdlib subpackages, so the blast radius is wide.

## Environment

pyscn `v1.30.1` (commit `2c1ff91`)

## Found via

FP-audit skill in repo `dajiaji/pyseto@628d6a34e0ede88abcb8234108272eda6745468d`

7487:
## Summary

`GroupNamespaceImports` collapsed `alias.Member` references only for `import mod as alias`. A binding created by `from pkg import mod` is the same kind of namespace, but `collectImports` never registered it in `namespaceAliases`, so every `mod.Member` reference became its own CBO edge (3x the coupling for identical class bodies in the issue's repro; `pyseto`'s `V1Public` reported at 12/high).

## Type of Change

## Related Issues
Fixes #732

## Changes
- `internal/analyzer/cbo.go`: the `NodeImportFrom` branch of `collectImports` now registers `module.name -> binding` in `namespaceAliases`, so `canonicalDependencyName` collapses `binding.Member` exactly as it does for `import pkg.mod as mod`. No change to `canonicalDependencyName` itself, and plain `import mod` is untouched (`TestCBOAnalyzer_NonAliasedModuleMembersAreNotGrouped` still holds).
- On the module-vs-class ambiguity raised in the issue: the import statement alone can't tell them apart, so the binding is registered only when its resolved name reads as a module under the PEP 8 heuristic the analyzer already uses (`looksLikeClassReference`, incl. `knownLowercaseClasses`). A CapWords binding like `from models import Widget` is left alone, so `Widget.Inner` in a type hint keeps its own identity; `from datetime import datetime` stays a class. I went with this instead of a usage-shape pass because CapWords bindings already collapse to the binding via `callDependencyName` (`Widget.create()` -> `Widget`), so the only `from`-import bindings that ever reach `canonicalDependencyName` as `x.Member` are snake_case ones — and for those, collapsing to `x` gives one edge whether `x` turns out to be a module or a lowercase class.
- Extracted `looksLikeClassName(resolved)` from `looksLikeClassReference` so the heuristic can run while imports are still being collected (the analyzer's `importedNames` map isn't populated yet at that point).
- `internal/analyzer/cbo_test.go`: regression test with the issue's repro, the `from pkg import _enums as reflection` form, and the CapWords guard; also asserts the per-member edges come back with `GroupNamespaceImports = false`. Fails on `main`, passes with the fix.

Issue repro on the built binary:

| class | before | after |
| --- | --- | --- |
| `FromImportUser` | 3: `serialization.Encoding`, `serialization.NoEncryption`, `serialization.PrivateFormat` | 1: `serialization` |
| `AliasImportUser` | 1: `serialization` | 1: `serialization` |

`dajiaji/pyseto@628d6a3`, `pyseto/versions/v1.py`: `V1Public` 12 (high) -> 7 (medium), `V1Local` 5 -> 4.

## Testing
- [x] `make test` passes locally (`go test ./...`)
- [x] `make lint` passes locally (`go vet` + `golangci-lint` v2.4.0 via `go run`: 0 issues)

## Documentation
- [x] Updated godoc comments


DaisukeYoda (approved):
LGTM. Registering the from-import binding at collection time so that the existing canonicalization path handles it unchanged is the right shape for this fix, and extracting the pure name heuristic keeps the module-vs-class judgement in one place. The test covering both grouping modes and the CapWords guard makes the intended boundary clear.

PR #752, opened 2026-09-08
feat: write every requested analyze report

### LINKED ISSUE #739: Allow multiple output formats in a single analyze run
`pyscn analyze` accepts one output format per invocation, so users who want both JSON (for CI parsing) and HTML (for reading) run the full analysis twice (example: DOI-USGS/dataretrieval-python `.github/workflows/code-health.yml`).

Expected: allow repeated/comma-separated formats, e.g. `pyscn analyze --json --html --no-open src`, running the analysis once and writing every requested report.

7487:
## Summary
`pyscn analyze` accepted a single format flag per invocation, so CI jobs that want JSON for parsing and HTML for reading ran the full analysis twice. Every selected format is now written from one run.

## Type of Change

## Related Issues
Closes #739

## Changes
- `service.OutputFormatResolver.DetermineAnalyzeReports` returns every requested `ReportFormat` in flag order (HTML when none) instead of erroring on more than one; the four-flag `Determine` keeps its single-format contract.
- `generateOutput` loops over the reports and writes each through a new `writeReport`. Timestamped files from one run share the stem, so `--json --html` yields `analyze_20260908_084103.json` and `analyze_20260908_084103.html`.
- `--output` keeps its single-file contract: it is rejected with a clear error when several formats are requested (one path cannot hold two reports, and concatenating them on stdout is not useful).
- The standalone community JSON case (`--json --select communities`) is decided per report, so `--json --html --select communities` still writes standalone JSON and a unified HTML report.
- Docs: `website/docs/cli/analyze.md`, README, and the command's `--help` examples.

```
$ pyscn analyze --json --html --text --no-open --select complexity src/
📊 Unified HTML report generated: .../.pyscn/reports/analyze_20260908_084103.html
📊 Unified JSON report generated: .../.pyscn/reports/analyze_20260908_084103.json
📊 Unified TEXT report generated: .../.pyscn/reports/analyze_20260908_084103.txt

$ pyscn analyze --json --html --output x.json src/
Error: --output takes a single report format; drop it to write all 2 reports to the reports directory
```

## Testing
- [x] `go test ./cmd/pyscn ./service` passes locally
- [x] `go vet ./cmd/pyscn ./service` passes locally (my local golangci-lint is v1 and refuses the repo's v2 config, so relying on CI for `make lint`)
- [x] Added tests: `TestAnalyzeCommandMultipleFormats` (end-to-end: two reports with a shared stem, `--output` rejected), `TestOutputFormatResolver_DeterminesAnalyzeReports` (flag-order list), and updated the text-format and community-JSON tests for the new signatures
- [x] Manual testing completed (above)

## Documentation
- [x] Updated godoc comments
- [x] Updated README or other docs (if needed)


7487:
Merged `main` in to clear the conflict. The only clash was in `generateOutput`: #750 moved `response.Version = version.Version` into `AnalyzeUseCase`, while this branch had rewritten that function into a per-report loop. Resolution keeps the loop and drops the assignment (and the now-unused `internal/version` import), so version/schema_version still come from the use case for every format written.

CI is green on the merge commit.

DaisukeYoda:
This reads well: the format flags now resolve to an ordered list through a single helper shared with the legacy resolver, the per-report write is cleanly separated, and the standalone community JSON decision correctly follows the report being written rather than the flag state.

One request: the rejection of `--output` combined with several formats happens only after the whole analysis has finished. Since it is purely a flag conflict, it should fail before any analysis starts, alongside the existing `--select` and `--min-severity` validation, so a user does not wait through a long run just to hit a usage error.

7487:
Good catch — moved in 49bf71e. `reportFormats()` only reads the format flags, so the check sits in `runAnalyze` right after the `--select` and `--min-severity` validation and returns before the use case is built. `generateOutput` just writes the reports now.

Extended `TestAnalyzeCommandMultipleFormats` to pin the ordering rather than just the error: it runs `--json --html --output -` against a path that does not exist and asserts the error is about `--output`. With the check back in `generateOutput` that test fails with `FILE_NOT_FOUND` instead, which is exactly the long-run-then-usage-error case you described.

`go test ./cmd/pyscn ./service` passes.

DaisukeYoda (approved):
Thanks!

PR #778, opened 2026-09-14
feat: add .gitignore to output directory

### LINKED ISSUE #775: [FEATURE] add `.gitignore` file to `.pyscn` folder
## Is your feature request related to a problem? Please describe.

running pyscn on a new codebase creates the `.pyscn` folder in the project root.
This folder should not be committed, so there's the need to add it to `.gitignore`.
Modern devtools solve this by creating a `.gitignore` file inside their cache folders with the `*` content.

## Describe the solution you'd like

Add a `.gitignore` file in the `.pyscn` folder with the `*` content, so that the folder is automatically ignored by git.

## Describe alternatives you've considered

None

## Example use case

-

## Additional context

This will avoid having to add the folder to the root .gitignore file, and lower the risk of polluting the git history if somebody forgets to add the ignore manually.

sanzoghenzo:
## Summary

Creates a .gitignore file in the output folder with `*\n` as content.

## Type of Change
- [ ] Bug fix (non-breaking change)
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring

## Related Issues

Closes #775

## Changes

- added the function `ensureOutputGitignore` to `utils.go`
- `generateOutputFilePath` calls the new function, displaying a warning if there are any error in the .gitignore creation. this way, the output file generation continues to work.

## Testing
- [x] `make test` passes locally - this is true for the `utils.go` module, there are many failing tests in the other modules.
- [x] `make lint` passes locally

## Documentation
- [ ] Updated godoc comments

## Additional Context

I don't have make on my windows machine, so I just run the commands manually.
Nowadays, I'm using task or mise to create tasks for my projects 😉


DaisukeYoda (changes requested):
@sanzoghenzo 

Thanks for the PR! One blocking issue: the .gitignore is also written into a user-configured output.directory, so pointing it at . would ignore the whole repository. Please create it only for the default .pyscn location, ideally at .pyscn/.gitignore.

(I didn't know Windows doesn't support make. Maybe we should add a mise config.)

sanzoghenzo:
> pointing it at . would ignore the whole repository.

Oh, I didn't think about that edge case. And it also doesn't make sense to add it to a path if it is outside of the project (like `../pyscn-analysis`)

> Please create it only for the default .pyscn location, ideally at .pyscn/.gitignore.

This will complicate a bit the interface, since there's no concept of "default directory" right now.
I'll think about (or better yet I will do a grinding session with the AI) whenever possible

DaisukeYoda:
@sanzoghenzo 

How about a simpler rule: write the .gitignore only when pyscn creates the directory itself. Existing directories (including `.`) are left untouched, and no "default directory" concept is needed.
```go
    _, statErr := os.Stat(outputDir)
    created := os.IsNotExist(statErr)

    if mkErr := os.MkdirAll(outputDir, 0o755); mkErr != nil { ... }

    if created {
        if err := ensureOutputGitignore(outputDir); err != nil { ... }
    }
```
`.pyscn/reports/.gitignore` is fine as the location. The rest of the PR can stay as is.

sanzoghenzo:
Thanks for the tip @DaisukeYoda !

Done.

DaisukeYoda (approved):
I pushed a gofmt fix to your branch. Merging, thanks!

